PR #69, opened 2024-11-12
some doc changes

omar-polo:
This is just a first pass over the documentation, fixing what `mandoc -Tlint` highlights.  In a few cases, some sentences were slightly changed (like for `plakar diff` usage which was wrong.)

small bonus: I fixed some typos with the aid of codespell.


poolpOrg:
Thanks !

PR #123, opened 2024-11-21
86 fix fuse

poolpOrg:


plakarfs/dir.go
```diff
@@ -0,0 +1,163 @@
...
+func (d *Dir) Lookup(ctx context.Context, name string) (fs.Node, error) {
+	if d.name == "/" {
+		return &Dir{parent: d, name: name, repo: d.repo}, nil
+	} else if d.parent.name == "/" {
+		return &Dir{parent: d, name: name}, nil
+	} else {
+		cleanpath := filepath.Clean(d.fullpath + "/" + name)
+		fi, err := d.vfs.Stat(cleanpath)
+		if err != nil {
+			return nil, err
```
omar-polo, line 103: this makes accessing non-existant files returning an Input/output error instead of ENOENT.  I'm not sure what errors we should return thought, `os.ErrNotExist` still yields a "input/output error" in the end.

omar-polo (approved):
reads fine and works for me on linux.

plakar crashes when I umount that directory, but the stacktrace points inside the fuse library.  i'll look into that.

it could be helpful to maybe print something when we successfully mount the fuse fs.

plakarfs/dir.go
omar-polo: it is `syscalls.ENOENT`

PR #124, opened 2024-11-21
make rebuildState public so it can be called externally, most notably

poolpOrg:
by any long-standing command to catch up on new snapshots that happened since the command was launched


omar-polo (approved):
that's really nice!  I can refresh and see new snapshots on the UI

PR #133, opened 2024-11-21
don't check update at every invocation

omar-polo:
Instead of checking for updates at every invocation, try to check at most daily.

Not a fan of the stat + open TOCTOU, but I don't know a better way around.  We could `open(O_RDWR|O_CREATE)` but then we don't know if we're the ones that created the file or it was created by someone else very recently.

This means that concurrent plakar processes with the wrong timing could all end up checking for update, but at least it's only an issue for concurrent invocations.

I've also tried to simplify CheckUpdate a bit.  it's probably easier to read the two commits one by one instead of the cumulative diff.


poolpOrg (approved):
LGTM

PR #141, opened 2024-11-22
errors are no longer stored as an array below DirEntry, instead

poolpOrg:
they are stored unitarily in packfiles and DirEntry can use the error entry checksum as a list head.

### DIFF
--- a/repository/state/state.go
+++ b/repository/state/state.go
@@ -67,11 +67,14 @@ type State struct {
 	muSnapshots sync.Mutex
 	Snapshots   map[uint64]Location
 
-	muDeletedSnapshots sync.Mutex
-	DeletedSnapshots   map[uint64]time.Time
-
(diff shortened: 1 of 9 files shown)


repository/state/state.go
```diff
@@ -264,7 +268,17 @@ func (st *State) Merge(stateID objects.Checksum, deltaState *State) {
 			subpart.Length,
 		)
 	}
-	deltaState.muSignatures.Unlock()
```
omar-polo, line 267: is it on purpose?

poolpOrg: absolutely not, I created the pull request on the wrong commit, this got fixed in a commit I didn't include in the PR to split the diff. sorry, closing the PR and will reopen with the this fix and the int vs uint64 fix that Copilot choked upon.

PR #173, opened 2024-11-27
api: returns json errors for validation failures

omar-polo:


omar-polo:
While here I've also taken the time to introduce some helpers to avoid the repetitions.

In a few cases I changed from `mux.Vars` to accessing the query string as offset and length are query parameters, not path component.

poolpOrg:
Since I accidentally set it ready for review, i'm unsure if you're expecting a review at this point or not, let me know when you want me to have a look :-)

omar-polo:
It's ready to be reviewed :)

poolpOrg (approved):

PR #175, opened 2024-11-27
remove metadata object, it'll be replaced by Omar's indexes

poolpOrg:
field Metadata still exists in snapshot Header, but will hold the checksum of an Index used to track metadatas


omar-polo (approved):
agree on the plan

PR #183, opened 2024-11-28
Set HTML5 doctype when rendering a file with highlight

brmzkw:


omar-polo (changes requested):
isn't there the wrong place?  If we loop multiple times we risk to write a doctype in the middle of documents.  I believe it should be around line 83 after the content-type

```
w.Header().Set("Content-Type", "text/html")  // after here
w.WriteHeader(http.StatusOK)
if _, err := w.Write([]bytes(""<!doctype html>")); err != nil {
	return err
}
```

omar-polo (approved):
LGTM

PR #214, opened 2024-12-10
provide flags field for object and chunks in anticipation for future …

poolpOrg:
…optimizations


omar-polo (approved):
don't see anything wrong with it

PR #218, opened 2024-12-13
Gchehade/vfs to fs interface

poolpOrg:


omar-polo (approved):
Very happy to see this!

I think we can optimize a few bits later but it's going to simplify so much code...

Issue #92, opened 2024-11-18
errors package


poolpOrg:
Plakar currently propagates system errors or generates fmt.Errorf() for custom errors.

An errors package has been added to the project and should provide custom errors we can rely upon in the whole tree.

poolpOrg:
This was discussed with @omar-polo we'll have per-package errors rather than a global errors package.

PR #220, opened 2024-12-13
improve error handling (for the apis but not only)

omar-polo:
This is linked to #92: only that instead of moving the errors to a separate package, I just started to expose them in the package they belong.  I think it's better to have `repository.ErrPackfileNotFound` rather than `errors.ErrPackfileNotFound` (not to mention the potential name clash with the stdlib' errors.)

I think that a follow-up to this could be to wrap the low level errors (in particular `repository.ErrBlobNotFound`) in the other packages, so that for e.g. we can expose `snapshot.ErrNotfound` instead of returning blob not found when we fail to load a state.

This also teaches the API layer to return a 404 on these errors.  The list could be extended.

Finally, i've removed a few `return err` in the APIs where they could cause issues.  We return an error in the handlers to have an automatic http response, but if we fail midway thru we shouldn't return err (since the error handler will try to write another set of headers after the truncated relpy.)


poolpOrg (approved):
LGTM

PR #221, opened 2024-12-14
fix directory errors

omar-polo:
This removes some leftovers from my errors-as-index changes and fixes the error accounting per-directory.  I'm not sure it's correct to update `Below.Errors` but the UI now shows both the global number of errors and the per-directory ones, so it's a success I guess :)


caching/scan.go
```diff
@@ -113,86 +113,6 @@ type ErrorEntry struct {
 	Error    string
 }
 
-func (c *ScanCache) EnumerateErrorsWithinDirectory(directory string, reverse bool) (<-chan ErrorEntry, error) {
```
poolpOrg, line 116: beautiful

snapshot/vfs/dirEntry.go
```diff
@@ -31,9 +31,6 @@ type DirEntry struct {
 	Classifications []Classification `msgpack:"classifications,omitempty" json:"classifications"`
 	CustomMetadata  []CustomMetadata `msgpack:"custom_metadata,omitempty" json:"custom_metadata"`
 	Tags            []string         `msgpack:"tags,omitempty" json:"tags"`
-
-	/* Errors */
-	Errors *objects.Checksum `msgpack:"errors,omitempty" json:"errors,omitempty"`
```
poolpOrg, line 36: I'm not 100% convinced about this one, needs to be discussed:

I understand why you do it: having the errors root as a btree in the snapshot header and an O(log n) complexity to locate the scan startpoint makes it unnecessary to track scan startpoints for nodes of the vfs, performances are good enough to always search from the root of the btree.

The counter argument is that several structures hold a bit of redundancy in case of a repository corruption, so that if shit ever hits the fan, we're not helpless at trying to recover. For instance, objects structures hold chunks checksums, so fileEntry could store the object checksum... but it stores the object structure so it has chunks checksum redundancy. In this case, if we stored the scan start point in the dirEntry, then in case of a corrupted packfile affecting the root of the btree, we might still be able to locate subparts.

What do you think ?

omar-polo: I've thought a bit about it.  Usually I'm not for duplicating information, but in this case this is all supposed to be immutable, and it could help both for recovery and to make sure we don't introduce errors (for e.g. a `plakar verify` could check this as well) so I'm convinced.

It'd be weird for the b-tree api to return the node itself, but not too bad.  I'll think of a nice way to export that information.

poolpOrg (approved):
LGTM

PR #223, opened 2024-12-14
simplify storage layer by returning an interface

poolpOrg:
remove bytes accounting, needs to be done at the repository level remove opt_stats, this will be turned into a prom exporter in agent mode


omar-polo (approved):
LGTM

PR #229, opened 2024-12-16
use go-cdc-chunkers from PlakarKorp

omar-polo:
noticed after the typo on the readme; we were still using go-cdc-chunkers from PlakarLabs.


poolpOrg (approved):
LGTM

PR #232, opened 2024-12-16
remove explicit comparer from ScanFrom

omar-polo:
it was a bad idea.  I've introduced because I thought it could have been handy for customizing the start of the iteration, but in practice it's just a footgun in the btree API.


poolpOrg (approved):
yeah LGTM

PR #234, opened 2024-12-18
Gchehade/streamable state

poolpOrg:


repository/state/state.go
```diff
@@ -141,37 +129,256 @@ func (st *State) getOrCreateIdForChecksum(checksum objects.Checksum) uint64 {
...
+	}
+	for _, checksum := range st.Metadata.Extends {
+		if _, err := w.Write(checksum[:]); err != nil {
+			return fmt.Errorf("failed to write checksum: %w", err)
+		}
+	}
+
+	if err := serializeMapping(w, st.DeletedSnapshots, func(key uint64) error {
+		return writeUint64(key)
+	}, func(value time.Time) error {
```
omar-polo, line 185: nit: could pass writeUint64 directly instead of wrapping it.

repository/state/state.go
```diff
@@ -141,37 +129,256 @@ func (st *State) getOrCreateIdForChecksum(checksum objects.Checksum) uint64 {
...
+		{"Snapshots", st.Snapshots},
+		{"Signatures", st.Signatures},
+		{"Errors", st.Errors},
 	}
 
-	st.rebuildChecksums()
+	for _, m := range mappings {
+		if err := serializeMapping(w, m.data, writeUint64, func(v Location) error {
+			return writeLocation(v)
+		}); err != nil {
```
omar-polo, line 214: nit: could pass `writeLocation` directly instead of wrapping it in a no-op.

omar-polo (approved):
two minor nitpicks; but looks good to me.  Love the change, looking forward to what it'll unlock :)

repository/state/state.go
poolpOrg: fixed

repository/state/state.go
poolpOrg: fixed !

PR #236, opened 2024-12-23
unplug snapshot statistics, part of them were moved to vfs summary re…

poolpOrg:
…maining bits

needs to be reimplemented out of snapshots


omar-polo (approved):
so the idea is to rely only on the VFS statistics instead of having this per-snapshot stats?  doesn't sound bad to me.

PR #237, opened 2024-12-23
fix cross-OS path handling

omar-polo:
filepath is a package that "implements utility routines for manipulating filename paths in a way compatible with the target operating system-defined file paths", as such it may use for e.g. forward or backward slashes depending on the OS.

path, instead, manipulates paths such as URLs that always use forward slashes, and it's what should be used when we handle backup paths internally.

lightly tested, and only on linux, but should help with Windows :)


poolpOrg (approved):
LGTM

PR #238, opened 2024-12-28
since big dir refacto, we no longer need to opendir/readdir as we

poolpOrg:
do not craft a children array, kill dead code


omar-polo (approved):
lovely

PR #239, opened 2024-12-28
when rootDir is a symlink, resolve prior to backup

poolpOrg:


omar-polo (approved):
i just wish we had realpath(3) available

LGTM

PR #240, opened 2024-12-29
235 de risk windows

poolpOrg:


omar-polo (approved):
Can't test but looks fine to me.  It's a bit annoying that we have to close the file before attempting a rename, but if that's how their filesystem works...

as discussed in private, there are still improvements possible, but i'm happy to know that now it at least works on windows :)

PR #242, opened 2024-12-29
btree improvements

omar-polo:
backport of several addition, improvements and bugfixes for our BTrees.

The new features (backward iterator, Update) will be needed for the vfs-on-btree PR that I'm going to send once this lands.

It's probably easier to read this commit by commit rather than with the full diff.


poolpOrg (approved):
LGTM

PR #245, opened 2024-12-29
when resolving a rootDir symlinks, push the origin link in the snapsh…

poolpOrg:
…ot as well


omar-polo (approved):
not sure if we have to do the same on windows too, but this looks good to me.

PR #246, opened 2025-01-06
make GetPackileBlob return a ReadSeeker

omar-polo:
This is another bit extracted from my vfs branch.  I'll need this to implement ReadSeeker on top of a fs.File for the API layer (in particular for the http.ServeContent in api_snapshot.go).

The main downside, and also why I'm extracting this into its own PR is because of the s/NewBuffer/NewReader/ change: a reader keeps all the data in memory while the reader is used, while the buffer removes it as it's read.  Didn't want to hide that in the much bigger vfs diff.


poolpOrg (approved):
LGTM

PR #247, opened 2025-01-06
remove dirty flag, it's no longer used since the switch to delta states

poolpOrg:
while at it, remove List entry as it's not used

### DIFF
--- a/repository/state/state.go
+++ b/repository/state/state.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"io"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/PlakarKorp/plakar/objects"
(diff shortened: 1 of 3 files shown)


repository/state/state.go
```diff
@@ -75,15 +74,10 @@ type State struct {
 	muErrors sync.Mutex
 	Errors   map[uint64]Location
 
-	muList sync.Mutex
-	List   map[uint64]Location
-
```
ptr-math, line 80: git add -p failure? Or it should be mentioned in the commit msg !

(A quick grep tells me it's the second option.)

ptr-math (changes requested):

ptr-math (approved):

mathieu-plak (approved):

PR #248, opened 2025-01-06
call snapshot.Close after each obtention of a snapshot instance

poolpOrg:


cmd/plakar/subcommands/cat/cat.go
```diff
@@ -59,6 +59,8 @@ func cmd_cat(ctx *context.Context, repo *repository.Repository, args []string) i
 
 	errors := 0
 	for offset, snap := range snapshots {
+		defer snap.Close()
```
omar-polo, line 62: defer in a loop isn't ideal.  maybe we'll want to split the body in its own function, but for now it should be ok.

cmd/plakar/subcommands/checksum/checksum.go
```diff
@@ -55,6 +55,8 @@ func cmd_checksum(ctx *context.Context, repo *repository.Repository, args []stri
 
 	errors := 0
 	for offset, snap := range snapshots {
+		defer snap.Close()
```
omar-polo, line 58: ditto

omar-polo (approved):

cmd/plakar/subcommands/cat/cat.go
poolpOrg: I very much agree but I also think for now it should be ok: we need to change the way the loop works so we don't have an array of *Snapshot, so we'll need a function as you suggest, but we need that done in a lot of subcommands so I'd rather do it as a separate diff.

mathieu-plak (approved):
While this makes me miss Destructors, it looks good (with the caveat noted by @omar-polo).

PR #249, opened 2025-01-06
rebuild the VFS on top of a BTree

omar-polo:
finally!

This could use some more testing: i did a few backup and restore but at the moment i can't try with the corpus on this laptop.  The sync is also something where I could use some help testing.

There are a few things that can be improved, some minor code repetition (like in the directory entries handling), but this has been dragged for a bit and maybe we can just merge this and improve in-tree.


poolpOrg (approved):
Generally where we're heading, get it in so we can fix any shortcoming as soon as possible

snapshot/backup.go
```diff
@@ -195,11 +207,20 @@ func (snap *Snapshot) Backup(scanDir string, options *BackupOptions) error {
 		maxConcurrency: make(chan bool, maxConcurrency),
 	}
 
-	ds := caching.DBStore[string, ErrorItem]{
+	filestore := caching.DBStore[string, vfs.Entry]{
+		Prefix: "__path__",
+		Cache:  sc2,
+	}
+	backupCtx.fileidx, err = btree.New(&filestore, vfs.PathCmp, 3)
```
mathieu-plak, line 214: ORDER seems a bit low, any rational for this value? Not a blocker but curious about it.

snapshot/check.go
```diff
@@ -16,106 +19,111 @@ type CheckOptions struct {
 	FastCheck      bool
 }
 
-func snapshotCheckPath(snap *Snapshot, fs *vfs.Filesystem, pathname string, opts *CheckOptions, concurrency chan bool, wg *sync.WaitGroup) (bool, error) {
+func snapshotCheckPath(snap *Snapshot, fsc *vfs.Filesystem, pathname string, opts *CheckOptions, concurrency chan bool, wg *sync.WaitGroup) (bool, error) {
```
mathieu-plak, line 22: small nit but you used vfs pretty much elsewhere so we should stick to it. It collissions with the package name tho so I don't know. But clearly we should _not_ use fs anywhere it's just too damn confusing.

snapshot/vfs/entry.go
```diff
@@ -0,0 +1,467 @@
...
+	Errors *objects.Checksum `msgpack:"errors,omitempty" json:"errors,omitempty"`
+}
+
+func NewEntry(parentPath string, record *importer.ScanRecord) *Entry {
+	target := ""
+	if record.Target != "" {
+		target = record.Target
+	}
+
+	ExtendedAttributes := make([]ExtendedAttribute, 0)
```
mathieu-plak, line 58: This could be len(record.ExtendedAttribute) to avoid any allocation down the line. NewEntry being in the hotpath I think it might matter.

snapshot/backup.go
```diff
@@ -29,8 +30,12 @@ type BackupContext struct {
 	imp            importer.Importer
 	sc             *caching.ScanCache
```
mathieu-plak, line 31: Don't think you even need this anymore.

snapshot/backup.go
```diff
@@ -368,38 +397,64 @@ func (snap *Snapshot) Backup(scanDir string, options *BackupOptions) error {
...
+	var previous *string
+	for fit.Next() {
+		path, _ := fit.Current()
+		if previous != nil {
+			if vfs.PathCmp(*previous, path) != -1 {
+				panic(fmt.Sprintf("unexpected ordering issue: %s vs %s", *previous, path))
+			}
+		}
+		previous = &path
+	}
```
mathieu-plak, line 415: That hunk is debugging only right? It's a good idea to have this, as part of wider testing indeed, but we should at least flag it with an XXX or w/e to remove it when we release.

snapshot/backup.go
```diff
@@ -40,13 +45,30 @@ type BackupOptions struct {
 	Excludes       []glob.Glob
 }
 
+func (bc *BackupContext) recordFile(file *importer.ScanRecord) error {
+	entry := vfs.NewEntry(filepath.Dir(file.Pathname), file)
+
+	path := file.Pathname
+	if path != "/" {
+		path = strings.TrimSuffix(path, "/")
+	}
```
mathieu-plak, line 54: That's the opposite of the old logic right? Why?

cmd/plakar/subcommands/archive/archive.go
```diff
@@ -142,105 +150,98 @@ func archiveTarball(snap *snapshot.Snapshot, out io.Writer, fs *vfs.Filesystem,
...
-			continue
+		header := &tar.Header{
+			Typeflag: tar.TypeReg,
+			Name:     filepath,
+			Size:     sb.Size(),
+			Mode:     int64(sb.Mode()),
+			ModTime:  sb.ModTime(),
+		}
+
+		if _, ok := fp.(fs.ReadDirFile); ok {
```
mathieu-plak, line 161: Not a huge fan of this, and this you have the stat already maybe just grab it from the Mode like we do elsewhere?

mathieu-plak (changes requested):

snapshot/backup.go
omar-polo: ops, good catch!  i've lowered it to 3 for testing, but it should be something higher.  for the errors i'm using 50, but that's a number put there without much investigation.  it could work here too however.

snapshot/check.go
omar-polo: ack, i'll change this to be `vfs`; i think fsc came from the name of the receiving pointer we use in the vfs package.

snapshot/backup.go
omar-polo: yeah, this is another debugging leftover.  i'll remove this, now we have btree.Verify which it's way better :)

cmd/plakar/subcommands/archive/archive.go
omar-polo: yeah, i agree it's better, thanks!  will fix below in archiveZip too.

snapshot/check.go
omar-polo: actually i can't use vfs here, it would shadow some vfs.* usage later in this function.

snapshot/backup.go
omar-polo: right, haven't spotted this, thanks!

snapshot/vfs/entry.go
omar-polo: fixed, thank you!

snapshot/backup.go
omar-polo: because in the btree we don't store directories with the final `/`; it simplifies a bit vfs.PathCmp and it's what path.Clean would do.

snapshot/check.go
mathieu-plak: Yeah that was my remark about collisions. fsc is fine-ish I guess. Let's go with this.

mathieu-plak (approved):
LGTM

PR #259, opened 2025-01-09
factor btree deserialization

omar-polo:
just something i noticed while i was fixing the error accounting.  this simplifies a bit and allows to more easily plug a version check in the future.

### DIFF
--- a/btree/btree.go
+++ b/btree/btree.go
@@ -2,7 +2,10 @@ package btree
 
 import (
 	"errors"
+	"io"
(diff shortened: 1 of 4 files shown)


omar-polo:
ah, while here also some GC of debugging code that shouldn't have been there.

poolpOrg (approved):
LGTM

ptr-math (approved):

mathieu-plak (approved):

PR #262, opened 2025-01-09
Stop using gorilla/mux in the API

brmzkw:
gorilla/mux is not maintained anymore, and the routing enhancements of go 1.22 (https://go.dev/blog/routing-enhancements) make it possible to only use the stdlib.

mux is still used in server/httpd/httpd.go so the dependency can't be removed yet.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -92,7 +92,8 @@ func SetupRoutes(server *http.ServeMux, repo *repository.Repository, token strin
 	authToken := TokenAuthMiddleware(token)
 	urlSigner := NewSnapshotReaderURLSigner(token)
 
-	server.Handle("/api/{path...}", APIView(func(w http.ResponseWriter, r *http.Request) error {
+	// Catch all API endpoint, called if no more specific API endpoint is found
+	server.Handle("/api/", APIView(func(w http.ResponseWriter, r *http.Request) error {
 		return &ApiError{
 			HttpCode: 404,
 			ErrCode:  "not-found",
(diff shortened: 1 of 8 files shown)


omar-polo (approved):
LGTM

api/api.go
```diff
@@ -55,92 +43,82 @@ func handleError(w http.ResponseWriter, err error) {
...
-	handle("/snapshot/vfs/children/{snapshot}:/", snapshotVFSChildren).Methods("GET")
-	handle("/snapshot/vfs/children/{snapshot}:{path:.+}/", snapshotVFSChildren).Methods("GET")
-	handle("/snapshot/vfs/children/{snapshot}:{path:.+}", snapshotVFSChildren).Methods("GET")
-
-	handle("/snapshot/vfs/errors/{snapshot}:/", snapshotVFSErrors).Methods("GET")
-	handle("/snapshot/vfs/errors/{snapshot}:{path:.+}/", snapshotVFSErrors).Methods("GET")
-	handle("/snapshot/vfs/errors/{snapshot}:{path:.+}", snapshotVFSErrors).Methods("GET")
-
-	return r
+	server.Handle("/api/{path...}", APIView(func(w http.ResponseWriter, r *http.Request) error {
```
omar-polo, line 95: micro nit: i won't use the path parameter when we're not going to use it, just `"/api/"` should do it here.  anyway, it looks fine to me, thanks!

omar-polo (approved):

api/api.go
brmzkw: done

poolpOrg (approved):
OK

PR #264, opened 2025-01-09
remove error from return args for GetSnapshots

sayoun:
as nothing in the function body can return an error


mathieu-plak:
Nice find but let's not merge this, State is about to move to on disk storage, and will be fallible once this is the case.

poolpOrg:
I was about to write "future implementations are likely to return errors" but @mathieu-plak beat me to it :-)

PR #268, opened 2025-01-10
Avoid generic interface{} and use generic types instead

brmzkw:
This is nitpicking. The types Item and Items used to return a single object and a list of objects in the API were holding a reference to an interface.

Now, instead, the structs take a generic type as a parameter.


omar-polo (approved):
lgtm

poolpOrg (approved):
yes !

PR #269, opened 2025-01-10
Gchehade/optimize distribution

poolpOrg:


snapshot/backup.go
```diff
@@ -583,6 +583,21 @@ func entropy(data []byte) (float64, [256]float64) {
...
+	if dataSize == 0 {
+		return [256]byte{}
+	}
+
+	var dist [256]byte
+	for i, f := range freq {
+		if f > 0 {
+			percentage := (f / float64(dataSize)) * 100
+			dist[i] = byte(percentage + 0.5)
+		}
```
omar-polo, line 596: dumb question time (apologize if it's stupid; i'll blame the context switch in that case :P) but: floats were in the 0...1 range or -1...1 range?  (i thought it was 0...1 but never checked)

omar-polo: oh ok, i got confused by the f > 0 check.

poolpOrg: if f == 0 then percentage is 0 and dist[i] is also zero, so this skips them

PR #270, opened 2025-01-10
vfs: provide a WalkDir implementation

omar-polo:
This works similarly to fs.WalkDir, but it's slightly more optimized for our use case: the std one has to read all the child entries in memory before processing them, while this implementation relies on the fact that the VFS layer returns the child entries already sorted.


poolpOrg (approved):
me likey

mathieu-plak:
Nice find! Indeed ReadDir does a sort, something I would have totally missed without profiling I think,

ptr-math (approved):

PR #271, opened 2025-01-10
state: Move indices to checksums.

mathieu-plak:
* For space optimization reasons, we were using uint64 indices in place of the real checksums. This commit moves us back to using checksum indices. This is indeed bad but it's in preparation for moving the state handling to disk rather than memory, where we don't really care about space but we'd like to get rid of the complexity of mapping checksums.

* Warning: this breaks the ondisk format of repository so be sure to trash your repo and caches before using it.

* This is _mostly_ mechanical but warrants a bit of care / proper review.

### DIFF
--- a/cmd/plakar/subcommands/info/info.go
+++ b/cmd/plakar/subcommands/info/info.go
@@ -296,42 +296,42 @@ func info_state(repo *repository.Repository, args []string) error {
 				}
 			}
 
-			for snapshotID, subpart := range st.Snapshots {
+			for snapshot, subpart := range st.Snapshots {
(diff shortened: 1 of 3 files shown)


omar-polo (approved):
makes sense to me given the objective.  with the tests fixed it's fine with me.

poolpOrg (changes requested):
tests do not pass, other than I agree with that as we discussed earlier and the diff LGTM

mathieu-plak:
Did I forget a git add? Yes of course! Now with adapted tests.

omar-polo (approved):
makes sense to me, ok

poolpOrg (approved):
GO GO GO

PR #272, opened 2025-01-10
update the btree even when the file was found in the cache

omar-polo:
when we scan the files we first put a Entry structure in the index and only later we consult the cache to see if we've already chunkified it.

The issues is that we have to update it even in the case the entry was found in the cache, otherwise we leave the Objects field to nil in the backup and the file content are not available.

### DIFF
--- a/snapshot/backup.go
+++ b/snapshot/backup.go
@@ -331,14 +331,6 @@ func (snap *Snapshot) Backup(scanDir string, options *BackupOptions) error {
 					fileEntry.AddClassification(result.Analyzer, result.Classes)
 				}
 
-				backupCtx.mufileidx.Lock()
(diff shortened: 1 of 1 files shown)


poolpOrg (approved):
been there :-)

mathieu-plak:
LGTM, honestly the code around this area could use some documentation (like the one you add), because I've tripped on this, and every time I spend time there I have to double read the logic to get things right. (And as evidenced by his message even gilles@ got bit by this in the past.

edit: To make it clear, I'm not saying this PR should do the documentation part, just that this is something we as a collective should think about doing!

omar-polo:
@mathieu-plak I agree.  While it's not incredibly convoluted, there are some delicate parts here and there and some more comments could help.

mathieu-plak (approved):

PR #274, opened 2025-01-13
use SplitN in ParseSnapshotID

omar-polo:
nit i spotted while investigating something else.  At first I thought we could `path.Clean` in here, but the second part after the `:` is not guaranteed to be a path (e.g. see `plakar find` whose manpage could use an update :P).


mathieu-plak (approved):
Reads fine to me. Again I guess this could use a bit of documentation :'.

poolpOrg (approved):
👍🏻

PR #275, opened 2025-01-13
fix some ls weirdnesses

### LINKED ISSUE #273: plakar ls prints the current directory as part of child
The following should not have printed private:

```
gilles@Gilless-Mac-mini plakar % ./plakar ls 33:/private          
2024-12-30T13:40:06Z drwxr-xr-x                      192 B /private
2024-12-30T13:40:08Z drwxr-xr-x                     2.5 kB /private/etc
```

in addition it should have printed the child node name, not the full path as follows:

```
gilles@Gilless-Mac-mini plakar % ./plakar ls 33:/private
2024-12-30T13:40:08Z drwxr-xr-x                     2.5 kB etc
```

I also found weird behaviours:

```
gilles@Gilless-Mac-mini plakar % ./plakar ls 33:/private 
2024-12-30T13:40:06Z drwxr-xr-x                      192 B /private
2024-12-30T13:40:08Z drwxr-xr-x                     2.5 kB /private/etc
gilles@Gilless-Mac-mini plakar % ./plakar ls 33:/private/
2024-12-30T13:40:06Z drwxr-xr-x                      192 B /private/
gilles@Gilless-Mac-mini plakar %
```

omar-polo:
Make sure the path is cleaned or the WalkDir callback breaks.

Do not print the named directory in its output.  In ls(1) you can achieve it only by passing -d, something which we don't have (and doesn't seem terribly useful in this context.)

Then, unless -recursive is passed, print only the entry name, not not its full path.

Fixes #273

### DIFF
--- a/cmd/plakar/subcommands/ls/ls.go
+++ b/cmd/plakar/subcommands/ls/ls.go
@@ -24,6 +24,7 @@ import (
 	"log"
 	"os"
 	"os/user"
+	"path"
 	"time"
 
 	"github.com/PlakarKorp/plakar/appcontext"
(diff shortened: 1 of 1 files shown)


poolpOrg (approved):
reads fine

mathieu-plak (approved):

PR #276, opened 2025-01-13
move the archive feature inside the snapshot

omar-polo:
will be soon used in the API layer as well, so prepare it.  a new feature, unused in the cli, is that now it's able to archive a given list of files/directories.


poolpOrg (approved):
Yes please

mathieu-plak (approved):
Makes sense

PR #277, opened 2025-01-13
backup: create BubbleTea output

brmzkw:


brmzkw:
The BubbleTea events processor outputs the number of directories and of files in the backup.

If we backup a directory, let's say /Users/niluje/xxx, the number of directories will include all the subdirectories of xxx, plus /, /Users and /Users/niluje, which might be surprising.

I should probably fix this before merging to main.

cmd/plakar/subcommands/backup/interactive.go
```diff
@@ -0,0 +1,157 @@
...
+	s += "\n"
+
+	s += fmt.Sprintf("      Files: %s %d", checkMark, m.countFilesOk)
+	if m.countFilesErrors > 0 {
+		s += fmt.Sprintf(" %s %d", crossMark, m.countFilesErrors)
+	}
+	s += "\n"
+
+	s += m.lastLog + "\n"
+	return s
```
omar-polo, line 107: nit: haven't checked how much garbage this produces, but you may consider using a `strings.Builder` here (which you can still feed to `fmt.Fprintf`)

omar-polo (approved):
to be fair, i haven't tested it yet and never used bubbletea before, but looks fine to me.

cmd/plakar/subcommands/backup/output.go
```diff
@@ -0,0 +1,26 @@
...
+	crossMark = lipgloss.NewStyle().Foreground(lipgloss.Color("#FF0000")).SetString("✘")
+)
+
+type eventsProcessor interface {
+	Close()
+}
+
+func startEventsProcessor(ctx *appcontext.AppContext, opt_quiet bool) eventsProcessor {
+	stdout := os.Stdout.Fd()
+	if !opt_quiet && term.IsTerminal(int(stdout)) {
```
mathieu-plak, line 22: We need a boolean dedicated to the fancy output. At least for debugging etc, we need the ability to have the verbose stdio output.

mathieu-plak (commented):
Looks good, just like op@  I'm new to all this bubble framework so I can't really comment on that part but it looks /sane/. I do have one objection with the PR though (see diff)

cmd/plakar/subcommands/backup/output.go
brmzkw: You can simply run `plakar backup | cat` to fallback on stdio

omar-polo: oh, good catch.  yeah, it would be nice, even though to be fair we could still use `plakar backup | cat`

mathieu-plak: Yeah sure I can cat it but I shouldn't have to dig the sources to find that out. The fancy output is akin to colors etc on standard shell tools so I don't think it's crazy to have an option for it.

cmd/plakar/subcommands/backup/interactive.go
brmzkw: done!

cmd/plakar/subcommands/backup/output.go
brmzkw: I added the option `plakar backup -stdio`

PR #278, opened 2025-01-13
simplify importer selection logic

omar-polo:
We always try first with what was given on the command line (or the current working directory); if that fails, maybe because it looks like like a URI but we don't support its "protocol", we still try to resolve it as a filesystem path.

### DIFF
--- a/cmd/plakar/subcommands/backup/backup.go
+++ b/cmd/plakar/subcommands/backup/backup.go
@@ -23,7 +23,7 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"path"
+	"path/filepath"
 	"strings"
 
 	"github.com/PlakarKorp/plakar/appcontext"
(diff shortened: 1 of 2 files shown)


omar-polo:
I tried a few combinations, but always with a local filesystem (don't have right now the way to test s3 or ftp).  I think the logic remains the same, but another pair of eyes wouldn't be bad :-)

omar-polo:
oh, ~a side effect that maybe i should point out is that we now always pass an absolute path to the filesystem importer~, and that `snapshot.Backup` now takes the importer as argument instead of opening it by itself, which may or may not be useful in the future.  I just disliked multiple `importer.NewImporter` calls.

brmzkw (approved):

mathieu-plak:
LGTM and approving (since there is no automerge on this one...), but I'd like Gilles to weigh in on that one as discussed on discord because we might want a broader fix here.

mathieu-plak (approved):

poolpOrg (approved):
OKI

PR #279, opened 2025-01-14
Fix extended attributes

brmzkw:
To initialize a slice, the capacity should use the third argument of make and not the second one.


poolpOrg (approved):
Nice catch

mathieu-plak (approved):
nice find

PR #280, opened 2025-01-14
Add omitempty to JSON outputs

brmzkw:
I'm not 100% sure this is enough. I believe we need to review all the API endpoints to make sure omitempty is set where it should, and not set where it shouldn't.

This will break the UI, but the branch jcastets/fix-omitempty is prepared to be merged.

### DIFF
--- a/objects/objects.go
+++ b/objects/objects.go
@@ -36,11 +36,11 @@ type Object struct {
 }
 
 // Return empty lists for nil slices.
-func (o Object) MarshalJSON() ([]byte, error) {
+func (o *Object) MarshalJSON() ([]byte, error) {
 	// Create an alias to avoid recursive MarshalJSON calls
 	type Alias Object
 
-	ret := Alias(o)
+	ret := (*Alias)(o)
 
 	if ret.Chunks == nil {
 		ret.Chunks = []Chunk{}
--- a/objects/objects.go
+++ b/objects/objects.go
@@ -35,6 +35,31 @@ type Object struct {
 	Flags           uint32           `msgpack:"flags" json:"flags"`
 }
 
+// Return empty lists for nil slices.
+func (o Object) MarshalJSON() ([]byte, error) {
+	// Create an alias to avoid recursive MarshalJSON calls
+	type Alias Object
+
+	ret := Alias(o)
+
+	if ret.Chunks == nil {
+		ret.Chunks = []Chunk{}
+	}
+	if ret.Classifications == nil {
+		ret.Classifications = []Classification{}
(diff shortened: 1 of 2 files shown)


objects/objects.go
poolpOrg, line 1: I think this is the wrong fix.

omitempty makes sense on msgpack for these but for json we should assume that only pointers can be nil and slices being nil is a nice of a missing initialization of some sort. The culprits would be these two functions, with the later being the one that actually affects you as what happens is that an empty array was serialized and saved, then upon reload since it was omitted you get a nil pointer and we don't properly reallocate the slice:

```go
func NewObject() *Object {
	return &Object{
		CustomMetadata: make([]CustomMetadata, 0),
	}
}

func NewObjectFromBytes(serialized []byte) (*Object, error) {
	var o Object
	if err := msgpack.Unmarshal(serialized, &o); err != nil {
		return nil, err
	}
	if o.CustomMetadata == nil {
		o.CustomMetadata = make([]CustomMetadata, 0)
	}
	if o.Tags == nil {
		o.Tags = make([]string, 0)
	}
	return &o, nil
}
```

objects/objects.go
```diff
@@ -35,6 +35,31 @@ type Object struct {
 	Flags           uint32           `msgpack:"flags" json:"flags"`
 }
 
+// Return empty lists for nil slices.
+func (o Object) MarshalJSON() ([]byte, error) {
+	// Create an alias to avoid recursive MarshalJSON calls
+	type Alias Object
+
+	ret := Alias(o)
```
omar-polo, line 43: I'd make this a pointer receiver to avoid having to copy the whole struct when serializing it; i.e.

```go
func (o *Object) MarshalJSON() ([]byte, error) {
        type Alias Object
        ret := (*Alias)(o)
        // ...
```

snapshot/vfs/entry.go
```diff
@@ -29,26 +30,55 @@ type Entry struct {
...
 	/* Errors */
 	Errors *objects.Checksum `msgpack:"errors,omitempty" json:"errors,omitempty"`
 }
 
+// Return empty lists for nil slices.
+func (e Entry) MarshalJSON() ([]byte, error) {
+	// Create an alias to avoid recursive MarshalJSON calls
+	type Alias Entry
+
+	ret := Alias(e)
```
omar-polo, line 58: ditto

omar-polo (approved):
left two nits; with that fixed it's ok for me (haven't tested the ui thought)

poolpOrg (approved):

PR #287, opened 2025-01-20
dep: Update lz4 to the latest version.

mathieu-plak:
* This version contains a fix for the compression of an empty stream. We had a workaround for it, but that workaround triggers another bug when used with State streaming (because we end up mismatching an Lz4Compression with both Write() and ReadFrom() implementations which is forbidden). By upgrading we can get rid of that workaround.


poolpOrg (approved):
yup yup

omar-polo (approved):

PR #295, opened 2025-01-21
rework the VFS so the BTree stores only the checksum of the entry

omar-polo:
For this, and to avoid changing too much the structure of backup.go, `btree.Persist` needs to learn how to convert the values: it allows to build the BTree as we're doing now, and deal with the checksums only when we have to persist the entries too.

### DIFF
--- a/btree/btree_test.go
+++ b/btree/btree_test.go
@@ -287,7 +287,7 @@ func TestPersist(t *testing.T) {
 	}
 
 	store2 := InMemoryStore[rune, int]{}
-	root, err := Persist(tree1, &store2)
+	root, err := Persist(tree1, &store2, func (e int) (int, error) { return e, nil })
 	if err != nil {
 		t.Fatalf("Failed to persist the tree: %v", err)
 	}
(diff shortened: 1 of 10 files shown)


omar-polo:
again, i've not really tested the sync; i still need to set up something, apologize.

omar-polo:
p.s. i'm open to suggestion for alternatives to `TYPE_INODE`.  Maybe `TYPE_ENTRY` or `TYPE_VFS_ENTRY`?

poolpOrg (approved):
OK

ptr-math (approved):
FTR I'm not a _huge_ fan of the conversion callback happening at persist time, but I really can't come up with anything better so ... I'll shut up and go on with my life.

OK;

mathieu-plak (approved):
FTR I'm not a huge fan of the conversion callback happening at persist time, but I really can't come up with anything better so ... I'll shut up and go on with my life.

OK;

omar-polo:
@mathieu-plak I tried at first to not do that, but it's harder since we have to update the structure (at least for now) and I couldn't come up with something better.

That mechanism will come in handy also for other stuff though, like i have a wip thing where i'm building a btree with empty structs as value that gets converted to the hash of the entries at persist time: also in that case I couldn't come up with something better.

Conceptually, it's like a map operation on the btree followed by the persist, only chained to avoid wasting time/space.

PR #296, opened 2025-01-21
Return HTTP/404 instead of 500 when browsing non-existing file

brmzkw:
I believe it's probably not the correct place where to make this check. Is there something better?


omar-polo:
I'd prefer to move this to `handleError` in `api.go`: the VFS now consistently returns `fs.ErrNotFound` for files/directories not found, or `fs.ErrInvalid` for invalid usage.  Thus, I'd prefer to map `ErrNotFound` to 404 and keep the implicit 500 for `ErrInvalid`.

omar-polo (approved):
yay, thanks!

poolpOrg (approved):
OK

PR #299, opened 2025-01-22
Create debug config for .vscode

brmzkw:
To use:

1. Set a breakpoint by clicking next to a line number

Screenshot 2025-01-22 at 11 54 18

2. On the debug menu, select your debug configuration. Two configurations exist: the first one runs plakar with the arguments provided, the second one starts the UI with `plakar ui -cors -addr localhost:30112 -no-spawn -no-auth`.

Screenshot 2025-01-22 at 11 54 29

3. Override the arguments, or press "enter" to keep the default

Screenshot 2025-01-22 at 11 54 40

4. When the breakpoint is reached, you get access to:
* top left: variables
* bottom left: call stack
* top right: actions (continue / step over / step into / restart / stop)
* bottom right: output

Screenshot 2025-01-22 at 11 54 52

If you’ve cloned all your repositories into a single parent directory (e.g., /users/xxx/PlakarKorp) and your .vscode folder is inside a subdirectory (e.g., /users/xxx/PlakarKorp/plakar/.vscode), debug configurations may not work because VS Code expects .vscode in the workspace root.

To setup a multi-root workspace:

1. Open VS Code in the plakar directory
2. Go to File > Add Folder to Workspace and select your other repos (e.g., plakar-ui).


omar-polo (approved):

poolpOrg (approved):
OK

PR #301, opened 2025-01-22
remove commit method from storage backends

sayoun:
doesn't seem to be used anywhere and not defined on the main storage.Store interface


poolpOrg (approved):
OK this is a leftover from previous repository format

omar-polo (approved):

PR #303, opened 2025-01-23
ondisk state v2

mathieu-plak:


snapshot/backup.go
```diff
@@ -835,17 +841,30 @@ func (snap *Snapshot) Commit() error {
...
 	return nil
 }
+
+func (snap *Snapshot) buildSerializedDeltaState() (io.Reader, error) {
+	pr, pw := io.Pipe()
+
+	/* By using a pipe and a goroutine we bound the max size in memory. */
+	go func() {
+		defer pw.Close()
+		snap.deltaState.SerializeToStream(pw)
```
poolpOrg, line 866: This should be a closewitherror to catch serialization issues

poolpOrg (changes requested):
added a comment

poolpOrg (approved):
OK

repository/state/state.go
```diff
@@ -17,24 +17,35 @@
...
 
 const VERSION = 100
 
+type EntryType uint8
+
+const (
+	ET_METADATA EntryType = iota
+	ET_LOCATIONS
+	ET_TIMESTAMP
+)
```
omar-polo, line 42: micro nit: for values that are going to hit the disk I think it'd be better to use explicit values instead of iota; makes it easier not to break the on-disk format inadvertently when adding values.

omar-polo (approved):
sorry for the delay!  lgtm

repository/state/state.go
poolpOrg: agreed, missed that

PR #307, opened 2025-01-24
fix checksum json unmarshalling

sayoun:
to avoid errors like 
```
json: cannot unmarshal string into Go value of type objects.Checksum
```

### DIFF
--- a/objects/objects.go
+++ b/objects/objects.go
@@ -1,6 +1,7 @@
 package objects
 
 import (
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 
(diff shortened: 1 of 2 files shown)


objects/objects.go
```diff
@@ -13,6 +14,20 @@ func (m Checksum) MarshalJSON() ([]byte, error) {
...
+	var s string
+	if err := json.Unmarshal(data, &s); err != nil {
+		return err
+	}
+
+	decoded, err := hex.DecodeString(s)
+	if err != nil {
+		return err
+	}
+	copy(m[:], decoded)
```
poolpOrg, line 27: You should check the length of decoded buffer to make sure it is equal to a [32]byte or error out: if decoded is not [32]byte it's not a legal input and we end up with two different situations, either it's < 32 bytes in which cases we have a broken checksum ended with 0s or it's > 32 bytes in which case the input will be truncated.

sayoun: @poolpOrg fixed in  https://github.com/PlakarKorp/plakar/pull/307/commits/cac081352e521e179370df91d2f2e4d6d27de997

poolpOrg (approved):
OK

omar-polo (approved):
lgtm

PR #311, opened 2025-01-25
Refactor fs backend a bit

ericfaurot:
Introduce a Buckets object that provides access to a bucket tree
Move generic os/fs helpers to utils.go
Simplify fs.go
Remove pathnames.go


storage/backends/fs/fs.go
```diff
@@ -74,75 +53,56 @@ func (repo *Repository) Location() string {
...
-		location = location[4:]
-	}
 
-	repo.root = location
+	repo.packfiles = NewBuckets(repo.Path("packfiles"))
+	repo.states = NewBuckets(repo.Path("states"))
 
-	configPath := filepath.Join(repo.root, "CONFIG")
-	rd, err := os.Open(configPath)
+	rd, err := os.Open(repo.Path("CONFIG"))
```
omar-polo, line 105: not a problem with your diff -- this issue is already present in the previous code, but we should eventually close this file.

storage/backends/fs/fs.go
```diff
@@ -173,54 +133,11 @@ func (repo *Repository) Configuration() storage.Configuration {
...
 }
 
 func (repo *Repository) GetPackfile(checksum objects.Checksum) (io.Reader, error) {
-	pathname := repo.PathPackfile(checksum)
-	if !strings.HasPrefix(pathname, repo.PathPackfiles()) {
-		return nil, fmt.Errorf("invalid path generated from checksum")
-	}
-
-	fp, err := os.Open(pathname)
+	fp, err := repo.packfiles.Open(checksum)
```
omar-polo, line 140: likewise here, the interface should allow to return a ReaderCloser -- but this may cause annoyance with the other backends.

I'm noting this here just to not forgot about it, I'll check how easily we could change the interface.

edit: TIL about `io.NopCloser` which should make this way easier.

omar-polo (approved):
LGTM, I like it!

one minor nit regarding the formatting (but apologise in advance) we usually use just one newline between functions.

I've also left a comment about the fact that we're leaking file descriptors (not a huge problem since the GC will close them eventually), but that will require changes in the interface that may be outside of the scope of this pull request.

storage/backends/fs/fs.go
ericfaurot: fixed

storage/backends/fs/fs.go
ericfaurot: fixed

omar-polo: thanks!

storage/backends/fs/fs.go
omar-polo: that's fine with me.  There's still the potential issue of not closing the file if we don't read it enough to hit EOF, but it's definitely better than what we have now.

poolpOrg (approved):
same comments as omar basically, this is good for now as it improves current state but we need to go revisit the possible fd leak

PR #313, opened 2025-01-25
Gchehade/spawn daemon agent

poolpOrg:
reduce upcoming pull requests by adding a `plakar agent` command that spawns the agent daemon as a noop and exports AppContext fields

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -1,0 +1,76 @@
+package agent
+
+import (
+	"fmt"
+	"net"
+	"os"
+
+	"github.com/PlakarKorp/plakar/appcontext"
+)
+
+type Daemon struct {
+	socketPath string
+	listener   net.Listener
+	ctx        *appcontext.AppContext
+}
+
+func NewDaemon(ctx *appcontext.AppContext, network string, address string) (*Daemon, error) {
+	if network != "unix" {
+		return nil, fmt.Errorf("unsupported network: %s", network)
+	}
+
+	d := &Daemon{
+		socketPath: address,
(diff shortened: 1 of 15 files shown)


omar-polo (approved):
LGTM; to be fair I guess we could delete now most of the appcontext tests since they're dummy.

agent/agent.go
```diff
@@ -0,0 +1,76 @@
...
+	if network != "unix" {
+		return nil, fmt.Errorf("unsupported network: %s", network)
+	}
+
+	d := &Daemon{
+		socketPath: address,
+		ctx:        ctx,
+	}
+
+	if _, err := os.Stat(d.socketPath); err == nil {
```
omar-polo, line 27: actually i forgot a small nitpick: I'm not sure about this extra stat, are you afraid of unlinking an existing file?

If I compare this with the "usual" `control.c` file, we just do an unlink (and error if it fails with something different than `ENOENT`), bind the socket and chmod it.

poolpOrg: yeah I don't want to unlink it if it exists and there's still an agent listening to it, I want to unlink it only if it exists but is a left-over from a previous run that crashed or was aborted.

poolpOrg:
> LGTM; to be fair I guess we could delete now most of the appcontext tests since they're dummy.

Yes, you're right

agent/agent.go
omar-polo: ok i see

mathieu-plak (approved):
Ok;

PR #317, opened 2025-01-27
Jcastets/linter

brmzkw:
The goal of this PR is not to fix the lint issues, but instead to add the linter to the CI to improve the quality of the future code.

We will probably need to spend some time to review what errors have been silenced and should have been handled instead.


caching/store.go
```diff
@@ -20,8 +20,8 @@ func (ds *DBStore[K, V]) Get(idx int) (*btree.Node[K, int, V], error) {
 		return nil, err
 	}
 	node := &btree.Node[K, int, V]{}
-	err = msgpack.Unmarshal(bytes, node)
-	return node , nil
+	_ = msgpack.Unmarshal(bytes, node)
+	return node, nil
```
omar-polo, line 24: this is the wrong fix, we should return err not hide it.

cmd/plakar/subcommands/exec/exec.go
```diff
@@ -90,7 +90,7 @@ func cmd_exec(ctx *appcontext.AppContext, repo *repository.Repository, args []st
 	}
 	go func() {
 		defer stdout.Close()
-		io.Copy(os.Stdout, stdout)
+		_, _ = io.Copy(os.Stdout, stdout)
```
omar-polo, line 93: don't wont to sound rude, but I hate this kind of things.  IMHO these constructs are less readable than the implicit discarding of the returned value.

(same for all the other places where this change was done -- I'm skeptical this really improves the quality of the codebase and/or avoids bugs.)

snapshot/archive.go
```diff
@@ -17,13 +17,13 @@ type ArchiveFormat = string
 
 const (
 	ArchiveTar     ArchiveFormat = "tar"
-	ArchiveTarball               = "tarball"
-	ArchiveZip                   = "zip"
+	ArchiveTarball ArchiveFormat = "tarball"
+	ArchiveZip     ArchiveFormat = "zip"
 )
```
omar-polo, line 22: also, fwiw, I *personally* don't like this; I'd prefer the impicit type rather than spell it everytime.  Not complaining this time, just raising my opinion to hear from what other think.

snapshot/archive.go
```diff
@@ -17,13 +17,13 @@ type ArchiveFormat = string
...
 
 var (
 	ErrInvalidArchiveFormat = errors.New("unknown archive format")
-	ErrNotADirectory = errors.New("is not a directory")
+	ErrNotADirectory        = errors.New("is not a directory")
```
omar-polo, line 26: actually this `ErrNotADirectory` is unused and can be removed -- I'll do it in a different PR to not clash with this one.

snapshot/exporter/fs/fs.go
```diff
@@ -70,12 +70,10 @@ func (p *FSExporter) StoreFile(pathname string, fp io.Reader) error {
...
-	if err := f.Sync(); err != nil {
-		//logging.Warn("sync failure: %s: %s", pathname, err)
-	}
-	if err := f.Close(); err != nil {
-		//logging.Warn("close failure: %s: %s", pathname, err)
-	}
+	_ = f.Sync()
+	//if err: logging.Warn("sync failure: %s: %s", pathname, err)
+	_ = f.Close()
+	//if err: logging.Warn("close failure: %s: %s", pathname, err)
```
omar-polo, line 76: meh; either fix the error handling or remove the comments.

caching/store.go
brmzkw: I agree, but I also believe this is probably something we should do in many other places.

snapshot/exporter/fs/fs.go
brmzkw: fixed

snapshot/archive.go
brmzkw: ok, I'm closing this conversation then!

brmzkw:
To recap the discussion of this morning on Discord, I'll do the following:

1/ update this PR to regroup fixes in different commits. For example, there will be one commit for `_ = flag.Parse`, one commit for another type of error, and so on.
2/ I will remove the command find which is causing linting issues, and will not be included in the beta.
3/ The PR can then be merged to main.
4/ Then, I'll create an issue with the following content: "The commit <xxx> silenced the linter. We need to review whether silencing was the right thing to do, if we should instead do something about it (eg. bubble up the error), or if we should update the linter configuration to ignore these errors and instead revert the commit". This issue will have to be reviewed to improve the current code.

caching/store.go
brmzkw: Resolving this comment, see https://github.com/PlakarKorp/plakar/pull/317#issuecomment-2619062380

snapshot/archive.go
brmzkw: Resolving this comment, see https://github.com/PlakarKorp/plakar/pull/317#issuecomment-2619062380

I personally don't mind about this syntax, but I also don't mind if we prefer to update the linter's configuration.

cmd/plakar/subcommands/exec/exec.go
brmzkw: See https://github.com/PlakarKorp/plakar/pull/317#issuecomment-2619062380

PR #318, opened 2025-01-27
Gchehade/cli commands prototype

poolpOrg:
most uninvasive diff possible to switch prototype of subcommands entrypoints from int to (int, error), we will be going through each individual command soon to cope with another branch work so no need to be too smart here.


brmzkw (approved):

poolpOrg:
> yay.
> 
> maybe we can also remove the various `ctx.GetLogger().Error` and rely on the caller to print out the error returned, but for the time being this LGTM!

yes definitely

PR #319, opened 2025-01-27
add storage backend plakard tests

sayoun:
I only tested the TCP client part, this will probably need to be refactored later to use a real plakard server instead of a mocked one.

before
```
	github.com/PlakarKorp/plakar/storage/backends/plakard		coverage: 0.0% of statements
```

after
```
ok  	github.com/PlakarKorp/plakar/storage/backends/plakard	0.006s	coverage: 53.9% of statements
```


omar-polo (approved):
+1 for eventually using the real server, but for the time being this LGTM!

poolpOrg (approved):

PR #320, opened 2025-01-28
when backing up a file, the scan directory needs to be reset to the f…

poolpOrg:
…ile parent directory


omar-polo (approved):
LGTM

mathieu-plak (approved):

omar-polo:
However I was wondering whether we should defer this even later.  The importer itself could adjust its root location and we could fill it in the snapshot after the scan phase.

brmzkw (approved):

PR #321, opened 2025-01-28
snap: Remove an useless copy in PutPackfile()

mathieu-plak:
* This used to be done to convert from the byte slice to the Checksum type. The repository.Checksum function now returns a Checksum so remove the now useless copy.


poolpOrg (approved):
yes

omar-polo (approved):
good catch

PR #322, opened 2025-01-28
Gchehade/add agent command proxy

poolpOrg:
this PR brings full local agent mode:
- bind a local agent that listens for requests
- handle requests by executing them in the agent process
- forward output and responses to the client (the CLI)
- stream events to the client

This is going to break multiple commands slightly as they need to redirect their output to a custom writer, but the fallouts can easily be fixed as a second PR following this one


appcontext/appcontext.go
```diff
@@ -69,10 +88,34 @@ func (c *AppContext) GetLogger() *logging.Logger {
 	return c.logger
 }
 
+func (c *AppContext) SetStdout(stdout io.Writer) {
```
ptr-math, line 91: We turned app context to a data structure a few commits ago, why make those private with do-nothing accessors?

ptr-math (approved):
One small nit to fix before merging. Otherwise LGTM. As discussed together this is a rough draft and needs to be refined in tree, but this can go in as is.

mathieu-plak (approved):

snapshot/backup.go
```diff
@@ -431,6 +438,13 @@ func (snap *Snapshot) Backup(scanDir string, imp importer.Importer, options *Bac
 	}
 
 	for fileiter.Next() {
+		select {
+		case <-snap.AppContext().GetContext().Done():
+			// Context canceled (e.g., due to client disconnect)
+			fmt.Println("Context canceled: client disconnected")
```
brmzkw, line 444: Can we log this message to os.stderr instead?

brmzkw (approved):

appcontext/appcontext.go
poolpOrg: that's so they don't get serialized

snapshot/backup.go
poolpOrg: removed, this was for debugging purpose.

agent/agent.go
```diff
@@ -57,20 +74,320 @@ func (d *Daemon) checkSocket() bool {
...
+			logger.EnableInfo()
+			clientContext.SetLogger(logger)
+
+			decoder := msgpack.NewDecoder(conn)
+
+			// Decode the client request
+			var request CommandRequest
+			if err := decoder.Decode(&request); err != nil {
+				if isDisconnectError(err) {
+					fmt.Println("Client disconnected during initial request")
```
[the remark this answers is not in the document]
poolpOrg: fixed

agent/agent.go
```diff
@@ -57,20 +74,320 @@ func (d *Daemon) checkSocket() bool {
...
+	cw.processFunc(string(p))
+	return len(p), nil
+}
+
+func isDisconnectError(err error) bool {
+	if err == io.EOF || err == io.ErrUnexpectedEOF {
+		return true
+	}
+	var netErr net.Error
+	return errors.As(err, &netErr) && !netErr.Temporary()
```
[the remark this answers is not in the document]
poolpOrg: fixed

appcontext/appcontext.go
poolpOrg: discussed over hackroom

PR #323, opened 2025-01-28
importer:fs: Implement a simple usr/grp name cache.

mathieu-plak:
* On nss enabled system, doing the username and groupname lookup is heavy (at least two syscalls) so let's use a local cache mapping uid/gid to names.

* This version is using an RWMutex making the diff slightly bigger than expected, but I think it's worth it, since the load is intrisically read-heavy.

* On Linux, this is a 200X speedup doing a pure fs import of the linux file source tree (from 40s down to 0.2s seconds run.)

### DIFF
--- a/snapshot/importer/fs/walkdir.go
+++ b/snapshot/importer/fs/walkdir.go
@@ -32,8 +32,15 @@ import (
 	"github.com/PlakarKorp/plakar/snapshot/importer"
 )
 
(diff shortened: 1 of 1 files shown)


omar-polo (approved):
love this, thank you!

poolpOrg (approved):
YUP YUP

PR #325, opened 2025-01-28
find is half-baked, lets remove it and reimplement it later when we

poolpOrg:
have better search capabilities


omar-polo (approved):
ack

mathieu-plak (commented):
Need to remove 'plakar/subcommands/help/docs/find.md' too. With that OK.

mathieu-plak (approved):

PR #326, opened 2025-01-28
remove snapshots code from s3 storage backend

sayoun:
it's not used, it doesn't match the storage interface, and no other backends use this


omar-polo (approved):
+1 for less dead code!  \o/

poolpOrg (approved):
KILL WITH FIRE

PR #328, opened 2025-01-28
initial implementation for a locate command to replace find

poolpOrg:
we have removed the find command until it has a better implementation backed by a better search mechanism, since we still need to be able to locate files, here's a locate command that allows searching snapshots for an entry or entry pattern.


cmd/plakar/subcommands/locate/locate.go
```diff
@@ -0,0 +1,101 @@
...
+		tmp := utils.LookupSnapshotByPrefix(repo, opt_snapshot)
+		snapshotIDs = append(snapshotIDs, tmp...)
+	} else {
+		tmp, err := repo.GetSnapshots()
+		if err != nil {
+			ctx.GetLogger().Error("%s: could not list snapshots: %s", flags.Name(), err)
+			return 1, err
+		}
+		snapshotIDs = append(snapshotIDs, tmp...)
+	}
```
omar-polo, line 55: nit: we can simplify this down to:

```go
var snapshotIDs [][]objects.Checksum
if opt_snapshot != "" {
	snapshotIDs = utils.LookupSnapshotByPrefix(repo, opt_snapshot)
} else {
	var err error
	snapshotIDs, err = repo.GetSnapshots()
	if err != nil {
		ctx.GetLogger().Error("...")
		return &, err
	}
}
```

which has the effect that it doesn't make me wonder what are we adding to where.

omar-polo (approved):
lgtm

cmd/plakar/subcommands/locate/locate.go
poolpOrg: indeed, pushed as a new commit

mathieu-plak:
@omar-polo Quick question, if we are not doing a pattern match, can't we take advantage of the btree to locate the file rather? Not sure the optim is worth it but... seeing us enumerate files over a btree to find a file irks me :D.

poolpOrg:
> @omar-polo Quick question, if we are not doing a pattern match, can't we take advantage of the btree to locate the file rather? Not sure the optim is worth it but... seeing us enumerate files over a btree to find a file irks me :D.

I dunno, `plakar locate` supports multiple parameters and unless we mark them as exact match or pattern, we don't know which are which before we attempt them. The only way to know would be to force escaping and pre-process args but then `plakar locate foobar.go` would become `plakar locate 'foobar\.go' and this feels less user-friendly.

Let's merge this and we can discuss optimizations

PR #331, opened 2025-01-29
163 prometheus exporter

poolpOrg:
this teaches agent to spawn a prometheus exporter and provides a prometheus.go file where we can add metrics we expose and functions to update these metrics.

while at it, make NewAgent() only create the Agent structure, not bind the socket, so that ListenAndServe is more robust and so failures to bind either the agent OR the prometheus exporter can be handled in a single place.


mathieu-plak (approved):
Seems straightforward. OK by me.

omar-polo (approved):

PR #336, opened 2025-01-29
fix race condition in subscription to events bus for the client:

poolpOrg:
if Execute() returns too fast, either due to invalid command or to a command returning too fast, the goroutine may have subscribed to the bus _after_ it's been closed leading to a deadlock on the synchronization channel. make sure to subscribe before entering the goroutine to ensure the subscription is valid after Execute.


omar-polo (approved):
lgtm

mathieu-plak (approved):

PR #338, opened 2025-01-29
remove keyring for now

poolpOrg:
they will be reintroduced in Q2


mathieu-plak (approved):

omar-polo (approved):
yay; small question: in backup.go we still have -identity, should that be removed as well?

brmzkw:
> yay; small question: in backup.go we still have -identity, should that be removed as well?

definitely!

brmzkw (approved):

PR #339, opened 2025-01-29
for now do not allow the agent socket to be configurable, this

poolpOrg:
raises too many questions wrt concurrency


omar-polo (approved):

omar-polo:
Just for the record: the way i've previously solved this issue is like what tmux does: take a lock on a file that's right next to the socket before attempting to link it.  this way you can deal with multiple clients trying to bind the socket and have only one actually registering it and the others just connecting.  Here's my implementation in amused if it can help: https://git.omarpolo.com/?action=blob&commit=911bd732a6cea22d332dad5a4e0adf4f0b19952c&file=ctl.c&folder=&path=amused.git#line935

brmzkw (approved):

PR #340, opened 2025-01-29
do not share cache between agent mode and agentless

poolpOrg:
this allows running agentless commands while the agent is running without interfering


mathieu-plak:
Why not but this needs a fair bit of warning, first run agentless might be surprising slow due to state synchronization.

mathieu-plak (approved):

brmzkw (approved):

PR #341, opened 2025-01-29
when I rewrote check to use the VFS, I mistakenly reported an error

poolpOrg:
on non-regular files instead of skipping them for chunking


mathieu-plak (approved):

omar-polo (approved):
ops, nice catch!

PR #349, opened 2025-01-30
state: Remove the old in memory State object.

mathieu-plak:
* Now that we have the leveldb backed one, remove this one as it's getting in our way to get new fields inside the structure.

* This also fixes plakar info state, not the best fix (we have to download repo info everytime) but it mirrors what was done before.

* Sadly I have to deactivate a bunch more tests.


repository/state/state.go
```diff
@@ -374,6 +383,22 @@ func (ls *LocalState) ListSnapshots() iter.Seq[objects.Checksum] {
 	}
 }
 
+func (ls *LocalState) ListObjectsOfType(Type packfile.Type) iter.Seq2[objects.Checksum, DeltaEntry] {
+	return func(yield func(objects.Checksum, DeltaEntry) bool) {
+		for csum, buf := range ls.cache.GetDeltasByType(packfile.TYPE_SNAPSHOT) {
+			de, err := DeltaEntryFromBytes(buf)
+
+			if err != nil {
```
poolpOrg, line 391: temporary ?

omar-polo (approved):

poolpOrg (approved):
OK

PR #350, opened 2025-01-30
do not allow creating repository with empty passphrase

poolpOrg:


brmzkw (approved):

omar-polo (approved):
let me nitpick on the wording; what about "can't encrypt the repository with an empty passphrase" or something along those lines?

still lgtm though; it should be disallowed unless `-no-encryption` is specified (and i'm unsure about assuming `-no-encryption` when an empty password is given)

poolpOrg (commented):
changed the wording to match your suggestion.

> and i'm unsure about assuming -no-encryption when an empty password is given

you only get prompted for a passphrase if the repo is set with encryption, so `-no-encryption` will not lead to that message

PR #351, opened 2025-01-30
rework the way the vfs btree is built

omar-polo:
The order of the inserts determines the shape of the btree: while it's always balanced and a different insertion order yields equivalent BTrees, the shape could be slightly different depending on where we split, and so the checksum could change.

Instead, insert the items in a stable order by deferring the BTree construction when we do the second pass to fix the statistics.


poolpOrg (approved):
LGTM

poolpOrg (commented):
iot looks even nicer on a second read <3

mathieu-plak (approved):

PR #352, opened 2025-01-30
add btreescan for profiling

omar-polo:
i've used this to spot possible regressions in the btree, but with a few tweaks it could be useful also to test improvements in the chucking and/or for the whole backup stage without having to remove the repository and cache.


poolpOrg (approved):
yes, lovely

poolpOrg (approved):

mathieu-plak (approved):

PR #353, opened 2025-01-30
when running a command in agent mode and there's no agent, we'd still rather have

poolpOrg:
commands succeed so warn that agent was down and fallback to agentless


omar-polo (approved):
YAY

mathieu-plak (approved):
Yes please!

PR #354, opened 2025-01-30
Add helpers to dump/initialize configuration to/from different format.

ericfaurot:
Simplify backend code accordingly.

### DIFF
--- a/storage/configuration.go
+++ b/storage/configuration.go
@@ -1,0 +1,141 @@
+/*
+ * Copyright (c) 2025 Eric Faurot <eric@faurot.net>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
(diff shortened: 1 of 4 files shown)


storage/configuration.go
```diff
@@ -0,0 +1,141 @@
...
+	"io"
+
+	"github.com/vmihailenco/msgpack/v5"
+
+	"github.com/PlakarKorp/plakar/compression"
+)
+
+func newConfigurationFromBytes(buffer []byte, format string) (*Configuration, error) {
+
+	if format == "auto" {
```
poolpOrg, line 32: not sure I'm convinced about automode, we'd rather be explicit in my opinion

storage/configuration.go
```diff
@@ -0,0 +1,141 @@
...
+	var unmarshal func(data []byte, v interface{}) error
+	gzip := false
+
+	switch format {
+	case "msgpack":
+		gzip = true
+		unmarshal = msgpack.Unmarshal
+	case "json":
+		unmarshal = json.Unmarshal
+	case "json+gz":
```
poolpOrg, line 54: is this used ? if you json+gz, i'm not sure why you would not use msgpack instead ?

poolpOrg (changes requested):
a few comments inlined

omar-polo:
With the recent work in progress by @poolpOrg this is going to be conflict, so this maybe should be closed and/or posticipated after the CONFIG changes.

poolpOrg:
Agreed, it not only conflicts but the storage layer now assumes the storage of a msgpack-serialized byte array, closing

PR #357, opened 2025-01-31
fix build on openbsd

ericfaurot:


poolpOrg (approved):
Ok

omar-polo (approved):
Thanks!

Can you please fix the other `mount_*.go` files as well?  The same mechanical diff should to work there as well.

PR #358, opened 2025-02-01
another stab at improving the manpages

omar-polo:


omar-polo:
for `plakar-agent.1` i ended up writing some chatgpt-style bullishit, sorry for that, but we'll improve that manpage later when we'll also have the config file.

There are still many things that I think could be improved, like consistently using real-scenario-style file paths in the examples (/var/www or /etc/passwd, stuff that's more meaningful than /path/to/file), consistently using the shell PS1 $ and sample output in the EXAMPLES section and improve the DIAGNOSTICS.

one step at the time however.

poolpOrg:
I fully agree.

Given the current state of man pages, I think this can go in regardless of any details, we can improve in tree.

poolpOrg (approved):
OK

mathieu-plak (approved):

PR #362, opened 2025-02-02
- server mode is agentless for the time being

poolpOrg:
- do not rebuild state in server mode, it is the client duty to rebuild on client side
- kill plakard server, we only support http servers for now


omar-polo (approved):
ok for me; can you also please adjust the manpage?

omar-polo:
ops; i forgot that the manpages are not yet committed; i'll fix them in my pr.

mathieu-plak (approved):

PR #365, opened 2025-02-03
rework encryption

poolpOrg:
long story short:

We don't need to store the encrypted master key in the repository config as if we store the KDF parameters and salt, we can derive the same master key from the passphrase and config parameters.

Adding a canary field with encrypted random data, we can rely on the GCM integrity check to validate that the passphrase supplied was the same as during repository configuration without leaking any useful informations.


omar-polo (approved):
not a cryptographer but lgtm

mathieu-plak (approved):

PR #373, opened 2025-02-03
improve the EXAMPLES section in various manpages

omar-polo:
- always use $ as "prompt" before the command examples
- try to use meaningful and well-known paths, like /etc/passwd or /var/www, instead of /path/to/dir or similar constructs
- try to be consistent with the paths, name of the tags and argument for example, both `plakar backup` and `plakar ls` use the tag named "daily-backup" in their examples section now
- finally, remove some redundant examples, we don't need to show every possible combination duplicating the DESCRIPTION section in the EXAMPLES section


poolpOrg (approved):
OK

brmzkw (approved):

PR #376, opened 2025-02-04
switch to multi-source snapshots, for now only one source allowed, provide a GetSource() method and hardcode GetSource(0) everywhere

poolpOrg:
switch to multi-source snapshots, make sure that only source 0 is used for now


omar-polo (approved):
YAY  \o/

brmzkw (approved):

PR #378, opened 2025-02-04
Fallback to -no-agent if agent is not available

brmzkw:
This replicates what has been mistakenly removed at 0f366cd95e31a


omar-polo (approved):
love it! \o/

poolpOrg (approved):
best diff ever

PR #383, opened 2025-02-04
switch all versions to versioning package

poolpOrg:


snapshot/header/header.go
```diff
@@ -63,26 +65,26 @@ func NewSource() *Source {
...
-	Name            string           `msgpack:"name" json:"name"`
-	Category        string           `msgpack:"category" json:"category"`
-	Environment     string           `msgpack:"environment" json:"environment"`
-	Perimeter       string           `msgpack:"perimeter" json:"perimeter"`
-	Classifications []Classification `msgpack:"classifications" json:"classifications"`
-	Tags            []string         `msgpack:"tags" json:"tags"`
-	Context         []KeyValue       `msgpack:"context" json:"context"`
-	Sources         []*Source        `msgpack:"sources" json:"sources"`
+	Identifier      objects.Checksum   `msgpack:"identifier" json:"identifier"`
+	Version         versioning.Version `msgpack:"version" json:"version"`
```
mathieu-plak, line 69: This should move to first place of the struct. I think we had that during the call, surely lost in all the staged changes.

mathieu-plak (commented):
One comment otherwise LGTM.

snapshot/header/header.go
poolpOrg: ok !

mathieu-plak (approved):

omar-polo (approved):
thank you!  The `vfs.Filesytem` needs still a bit f love but i can do a follow-up to this.

PR #384, opened 2025-02-04
state: Serialize length of entry alongside its type.

mathieu-plak:
* This allows us to skip unknown entry types (for forward compat')

PR #385, opened 2025-02-04
state: Add DeleteSnapshot() operation

mathieu-plak:
* /!\ Warning this is on top of #384, because this one might be a bit more controversial.

* let's you delete a snapshot by id by putting it into a dedicated list.

* While here do a bit of refactoring of the cache methods, as they grow
a bit too much for my liking.


poolpOrg (approved):
OK

omar-polo (approved):

PR #391, opened 2025-02-04
create: avoid infinite loop if stdin is a pipe

brmzkw:
it fixes `echo | plakar create` that was previously an infinite loop, but it's still not possible to provide the passphrase through a pipe.


poolpOrg (approved):
duh

omar-polo (approved):

PR #392, opened 2025-02-04
WIP: state: Record the list of packfiles created.

mathieu-plak:
* When building a delta state, record the packfiles that were created as part of the process.

* Warning this is based on #385 ...

* Putting in draft because I'm wondering if we should record the packfiles under the umbrella of the state id (so in leveldb it would be __packfile__:<state_id>:<csum>, and the payload would contain the state id), to efficiently retrieve it (for example in plakar info state f00b4r). This would need quite a bit of refactoring though (we only assign an ID to the state at serialization time right now).


poolpOrg (approved):
OK

PR #397, opened 2025-02-05
let repository provide a HMAC hasher preset with repository master key and a ChecksumHMAC helper

poolpOrg:
this will be used to generate HMAC for the storage file format and for privacy-friendly lookup keys

### DIFF
--- a/repository/repository.go
+++ b/repository/repository.go
@@ -2,7 +2,6 @@ package repository
 
 import (
 	"bytes"
-	"crypto/hmac"
 	"errors"
 	"hash"
 	"io"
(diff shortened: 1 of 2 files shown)


omar-polo (approved):

repository/repository.go
```diff
@@ -264,6 +264,25 @@ func (r *Repository) Checksum(data []byte) objects.Checksum {
...
+func (r *Repository) HMAC() hash.Hash {
+	return hashing.GetHMAC(r.Configuration().Hashing.Algorithm, r.AppContext().GetSecret())
+}
+
+func (r *Repository) ChecksumHMAC(data []byte) objects.Checksum {
+	hasher := r.HMAC()
+	hasher.Write(data)
+	result := hasher.Sum(nil)
+
+	if len(result) != 32 {
```
mathieu-plak, line 276: I don't think this is possible. HMAC returns the Sum from the underlying hashing algorithm, and they use arrays (not slices) as part of their APIs.

mathieu-plak (approved):
One small nit

PR #399, opened 2025-02-05
final Snapshot header version

poolpOrg:


mathieu-plak (approved):

poolpOrg:
This needs @omar-polo 's btree wrapper to be completed

omar-polo:
I've discussed this with @mathieu-plak this morning and we've decided to leave it as it is now.  It's in the wrong level to fix the thing, since we need the repository open for the VFS to work, and we can't do that at the snapshot.Header level.

omar-polo (approved):

PR #412, opened 2025-02-05
swap lookup key with a hmac before calling state lookups or repository lookups

poolpOrg:
if the resourceType is not snapshot, this allows building packfiles with field
checksum in the index set as the hmac and lets state and repo use correct keys


poolpOrg:
This needs careful reviewing.

The diff would be simpler if the state layer had access to the repository config and AppContext as the hmac computation could be hidden there as is done with repository layer, but this could be a future fix.

mathieu-plak (approved):

poolpOrg:
This was merged as a separate pull request

PR #414, opened 2025-02-06
instead of pushing a packfile with its checksum as identifier,

poolpOrg:
push a hmac !


poolpOrg:
nope, we can't do that, close

PR #419, opened 2025-02-06
Display a warning message when -cpu is <= 0

brmzkw:


cmd/plakar/plakar.go
```diff
@@ -150,6 +150,10 @@ func entryPoint() int {
 	}
 
 	// setup from default + override
+	if opt_cpuCount <= 0 {
+		fmt.Fprintf(os.Stderr, "%s: invalid -cpu value %d. Fallking back to 1.\n", flag.CommandLine.Name(), opt_cpuCount)
```
mathieu-plak, line 154: s/Fallking/Falling

omar-polo (approved):
ok with me with the typo fixed :P

cmd/plakar/plakar.go
```diff
@@ -150,6 +150,10 @@ func entryPoint() int {
...
+		opt_cpuCount = 1
```
mathieu-plak, line 155: We should fail rather than defaulting to 1. Just like the check below when you are over NumCPU.

mathieu-plak (approved):

omar-polo:
actually why not fail like done in the other cases?

cmd/plakar/plakar.go
brmzkw: fixed

PR #435, opened 2025-02-07
state: Save blob version when pushing them.

mathieu-plak:
* This way we don't have to go through the index to check a blob version.

* While here remove an old function that is unused anymore, reducing the number of way we push DeltaEntries to the state.


poolpOrg (approved):
Wonderful.

omar-polo (approved):

PR #437, opened 2025-02-07
s3: improve error messages on repository open failure

brmzkw:
We struggled to debug an error due to permissions because of the lack of context in errors messages. With this fix, it would have been easier.


omar-polo (approved):
yes!  i think we should do this more in general.

poolpOrg (approved):

PR #438, opened 2025-02-07
turn VisitDFS into an iterator

omar-polo:


poolpOrg (approved):

mathieu-plak (approved):
Love it. Pattern is way more clearer (since we are not mutating any of the internals anyway).

PR #449, opened 2025-02-08
fix build on openbsd, freebsd and windows

omar-polo:
while here rework the structure of the mount command so that only the Execute function is per-os; hopefully doing this will make it harder to unwillingly break the build on non-linux and non-mac OSes.


poolpOrg (approved):
OK !

mathieu-plak (approved):

PR #450, opened 2025-02-08
api: treat ErrPackfileNotFound as 500

omar-polo:
Unlike the snapshot not found case, which may be caused by a typo or a purposefully deleted snapshot, the packfile not found case is a sympthom of a worst problem: some kind of corruption in the repository.


omar-polo:
Actually we seem to return ErrPackfileNotFound even when the snapshot wasn't found (caught thanks to the tests!), so this needs some more work.

omar-polo:
made #1043 out of this, so this can be closed for now.

PR #451, opened 2025-02-08
improve packfile not found error reporting

omar-polo:
and while here make objects.Checksum more friendly in logs/error messages.

### DIFF
--- a/objects/objects.go
+++ b/objects/objects.go
@@ -20,6 +20,10 @@ func init() {
 
 type Checksum [32]byte
 
+func (m Checksum) String() string {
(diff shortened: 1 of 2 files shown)


omar-polo:
Actually we seem to return `ErrPackfileNotFound` even when the snapshot wasn't found (caught thanks to the tests!), so this needs some more work.

omar-polo:
Haven't realized that adding a `String()` method on `objects.Checksum has deeper consequences that I don't want to deal with right now, like breaking the way we construct paths given checksums.  Closing this for now.

PR #453, opened 2025-02-08
duplicate `info` as `diag` so we can simplify `info` for users,

poolpOrg:
and use `diag` for internal data structures diagnostics


cmd/plakar/subcommands/diag/plakar-diag.1
```diff
@@ -0,0 +1,73 @@
...
+.Bl -tag -width Ds
+.It 0
+Command completed successfully.
+.It >0
+An error occurred, such as an invalid snapshot or object ID, or a
+failure to retrieve the requested data.
+.El
+.Sh SEE ALSO
+.Xr plakar 1 ,
+.Xr plakar-snapshot 1
```
omar-polo, line 73: ops, this was wrong in the original file as well, it should be `plakar-info` here maybe, or just be removed.

omar-polo (approved):
makes sense to me; unsure about doing one `s/Info/Diag/g` over the structure names to reflect the command name, because i don't mind them using the Info prefix in this case.

cmd/plakar/subcommands/diag/plakar-diag.1
poolpOrg: ops ! will fix

mathieu-plak (approved):

PR #462, opened 2025-02-10
vfs: Export ResolveEntry()

mathieu-plak:
* This used to be private but now that we have a way of iterating over nodes (and as a result the checksum entries) it makes sense to have this public rather than coding it ad-hoc everywhere, especially since this does quite a bit of things.


poolpOrg (approved):

omar-polo (approved):
ok for me.  maybe i'd rename it at this point, this wasn't the best name i came up with, but that can be done as a follow-up.

PR #463, opened 2025-02-10
vfs: add another btree for extended attributes

omar-polo:
Since these can be varying in size (on darwin for example), we're tracking their content in an object like it were a file itself but to avoid cluttering the "real" VFS, we're keeping these entries in their own tree which the VFS package take care of.

The interface to handle xattrs will be improved in subsequent commits.


snapshot/importer/fs/walkdir_windows.go
poolpOrg, line 1: has to go through the same changes as walkdir.go so we don't have the symlinks processed twice and xattr sent after

poolpOrg (approved):
approved

mathieu-plak (approved):

PR #465, opened 2025-02-11
switch to AES256-GCM-SIV

poolpOrg:
Following crypto audit, it was suggested as an improvement to switch to AES256-GCM-SIV:

**We recommend switching to AES-GCM-SIV for chunk encryption.**

AES-GCM-SIV is a mode defined in [RFC 8452](https://www.rfc-editor.org/rfc/rfc8452.html) that does not rely on
randomness. It produces the nonce by computing a PRF over the message to
encrypt. It implies that encrypting the same message twice will produce
the same ciphertext. However, if each subkey encrypts a single chunk,
this is not an issue.

AES-GCM-SIV also prevents streaming of the data hashed, therefore the
whole chunk has to be stored in memory. Since data is already chunked to
be streamed, and chunks are of fixed, small size (64 KB), this is not an
issue.

That said, AES-GCM-SIV has less adoption than AES-GCM, and is not as
standardized as AES-GCM. AES-GCM is fine security-wise, switching to the
SIV mode would just eliminate one risk related to randomness. Depending
on the business requirements and client needs, AES-GCM may be preferable
(for example, if a FIPS standard is needed).

The most reliable implementation of AES-GCM-SIV is in [Google's Tink
package](https://pkg.go.dev/github.com/google/tink/go@v1.7.0/aead/subtle#AESGCMSIV).

Note that the Go language maintainers are planning to [add AES-GCM-SIV to x/crypto](https://github.com/golang/go/issues/54364#issuecomment-2460514455).

### DIFF
--- a/cmd/plakar/subcommands/info/repository.go
+++ b/cmd/plakar/subcommands/info/repository.go
@@ -54,8 +54,9 @@ func (cmd *InfoRepository) Execute(ctx *appcontext.AppContext, repo *repository.
 
 	if repo.Configuration().Encryption != nil {
 		fmt.Fprintln(ctx.Stdout, "Encryption:")
-		fmt.Fprintln(ctx.Stdout, " - Data Algorithm:", repo.Configuration().Encryption.DataAlgorithm)
-		fmt.Fprintln(ctx.Stdout, " - Subkey Algorithm:", repo.Configuration().Encryption.SubKeyAlgorithm)
+		fmt.Fprintln(ctx.Stdout, " - SubkeyAlgorithm:", repo.Configuration().Encryption.SubKeyAlgorithm)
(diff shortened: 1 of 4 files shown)


poolpOrg:
Note that our data encryption algorithm is configurable so we can introduce an AES-GCM variant for cases where it is required.

omar-polo (approved):
maybe we should update the library to https://github.com/tink-crypto/tink-go which seems to be the maintained one, otherwise looks good to me.

poolpOrg:
Had a look to tink-go but the API is different and diverges too much from the stdlib which we will switch to when FIPS standardizes SIV, we'll pass for the time being.

mathieu-plak (approved):

PR #478, opened 2025-02-11
rework and clean the scan interface a tiny bit

omar-polo:


snapshot/importer/importer.go
```diff
@@ -25,11 +25,12 @@ import (
...
 	"github.com/PlakarKorp/plakar/objects"
-	"github.com/vmihailenco/msgpack/v5"
 )
 
-type ScanResult interface {
-	scanResult()
+type ScanResult struct {
+	Record *ScanRecord
+	Error  *ScanError
+	Xattr  *ScanXattr
```
mathieu-plak, line 33: We don't need this and the struct anymore.

mathieu-plak (changes requested):
One comment inline, also btreescan is broken now!

snapshot/importer/importer.go
omar-polo: ops, right!

poolpOrg (approved):
ooouh yeah !

PR #479, opened 2025-02-11
main: Fixup cache creation.

mathieu-plak:
* When we fail to ExecuteRPC and fallback to agentless mode, we need to reopen the cache to actually use the agentless cache.


poolpOrg (approved):
Duh

omar-polo (approved):

PR #481, opened 2025-02-11
error log on fallback is WAY TOO VERBOSE, yeah we get it, you could not connect

poolpOrg:


mathieu-plak (approved):

mathieu-plak:
Intended to fix that with my previous one but somehow forgot about it.

brmzkw (approved):

PR #492, opened 2025-02-13
add context to importers for future uses of remote importers

poolpOrg:
in the future we will support remote importers, so we need their context


omar-polo (approved):
yeah, it would be nice to be able to add the metadata from the importer in the snapshot.

mathieu-plak (approved):

PR #496, opened 2025-02-13
factor out packfile creation

mathieu-plak:
This will be used by the cleanup function.


poolpOrg (approved):
OK

omar-polo (approved):
yeah, feels like the right level where to move those changes.  lgtm

PR #498, opened 2025-02-13
WIP: state: Add a new entry for configuration storage.

mathieu-plak:
* Since the state is the only mutable structure in our repository use it at our advantage to store configuration options for the repository.

* Unused for now but cleanup will use it soon.

It's in WIP because there is a caveat, since our state is an append only structure it makes deletion unpractical. I didn't want to have DeletedConfigurationEntries because it increases the complexity of it, but if we really want it we can. My best guess was to use the default value (empty slice) as a marker of deletion.

### DIFF
--- a/repository/state/state.go
+++ b/repository/state/state.go
@@ -29,6 +29,7 @@ import (
 	"github.com/PlakarKorp/plakar/resources"
 	"github.com/PlakarKorp/plakar/versioning"
 	"github.com/google/uuid"
+	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/vmihailenco/msgpack/v5"
 )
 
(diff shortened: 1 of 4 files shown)


repository/state/state.go
```diff
@@ -613,6 +697,49 @@ func (ls *LocalState) DeleteSnapshot(snapshotID objects.MAC) error {
...
+
+	return ls.insertOrUpdateConfiguration(ce)
+}
+
+// Internal function used by deserialization that only updates our local on
+// disk state if the provided configuration is more recent than the stored one
+func (ls *LocalState) insertOrUpdateConfiguration(ce ConfigurationEntry) error {
+	value, err := ls.cache.GetConfiguration(ce.Key)
+
+	if err != nil && err != leveldb.ErrNotFound {
```
mathieu-plak, line 719: Note to reviewers: Not sure about this pattern with the tri state check. I experimented with various versions (if/elseif etc) and settled on that one. Opinions?

poolpOrg: fine by me

omar-polo (approved):

poolpOrg (approved):
OK

PR #501, opened 2025-02-13
plakar: Fix implicit agentless error reporting.

mathieu-plak:
* Because a short variable form declaration was used the err variable was local and shadowed the global one, therefore when out of the inner scope the global err was nil and nothing was logged.


poolpOrg:
Oooooouh

poolpOrg (commented):
Ok

poolpOrg (approved):
Ok

omar-polo (approved):
oh i see!  good catch!

PR #503, opened 2025-02-14
rework sync to that it actually works

omar-polo:
The code in here rotted a little bit while we were doing big changes on the repository.  Make it work again by leveraging the new interfaces (for e.g. the new vfs), and by trying to sync *everything*.

### DIFF
--- a/cmd/plakar/subcommands/sync/sync.go
+++ b/cmd/plakar/subcommands/sync/sync.go
@@ -245,45 +245,103 @@ func (cmd *Sync) Execute(ctx *appcontext.AppContext, repo *repository.Repository
 	return 0, nil
 }
 
-func push(src *snapshot.Snapshot, dst *snapshot.Snapshot, mac objects.MAC, rtype resources.Type, data []byte) (bool, []byte, error) {
+func push(src *snapshot.Snapshot, dst *snapshot.Snapshot, mac objects.MAC, rtype resources.Type, data []byte) error {
+	if dst.BlobExists(rtype, mac) {
+		return nil
+	}
+	var err error
+	if data == nil {
+		data, err = src.GetBlob(rtype, mac)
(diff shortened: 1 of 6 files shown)


omar-polo:
I don't like how i'm dealing with the indexes, but for now this is maybe enough.  i need to move all this logic (which is sort-of duplicated with `diag contenttype`) into a different layer eventually.  i'm still not sure which.

cmd/plakar/subcommands/sync/sync.go
```diff
@@ -214,92 +219,262 @@ func (cmd *Sync) Execute(ctx *appcontext.AppContext, repo *repository.Repository
...
+		data, err = src.GetBlob(rtype, mac)
+		if err != nil {
+			return false, nil, err
+		}
+	}
+	return false, data, dst.PutBlob(rtype, mac, data)
+}
+
+func syncObject(src *snapshot.Snapshot, dst *snapshot.Snapshot, mac objects.MAC) error {
+	found, objbytes, err := push(src, dst, mac, resources.RT_OBJECT, nil)
```
poolpOrg, line 239: I think resources.RT_OBJECT should be pushed after the chunks that constitute it so that an object can't exist on the other end with an incomplete chunks set, no ?

poolpOrg (approved):
LGTM

cmd/plakar/subcommands/sync/sync.go
omar-polo: fixed

omar-polo:
As anticipated today in chat, I'm planning to 1) move this to another layer and 2) provide a better API to traverse (walk? :P) all the resources we need to sync/colour, but i still think that for the time being this should go in as-is since it fixes an issue.

mathieu-plak (approved):

PR #504, opened 2025-02-14
state: A few changes to how we key off DeltaEntry.

mathieu-plak:
* To be able to allow concurrent backups (possibily leading to same Blob in different packfiles) we need to add the Packfile id at the end of each blob Key, thereby allowing multiple Blob keys.

* This turns Retrieval into something a bit harder, and also slower but there is no other way to do this properly

* This gives us the ability to also not even consider any blob entry which packfile is not in our know state packfiles list, allowing atomical publication of entire packfiles' worth of blob.

* Key extraction, right now I just deserialize all payloads to lookup the info, I could do the same I was doing before, splitting on ':' and returning blobMAC + packfile, but this needs iter3 etc etc so I guess a bit of "back to the drawing board" is needed here. Still pushing this to get it out there.

* Point of attention for reviewers: The state.BlobExists() function, we "ignore" errors (as in they are not logged as proper errors but also make the blob as NotFound). It's my assumption that this is fine, since the only real errors we could have are a failing leveldb file access etc, and at that point any other call to leveldb will fail. We could log it but it'd make a lot of call site that do snap.BlobExists() a lot more complicated. Thoughts?


poolpOrg (approved):
OK

omar-polo (approved):

PR #507, opened 2025-02-16
PoolpOrg/utils-parseflags-date-resolution

poolpOrg:


cmd/plakar/utils/time.go
```diff
@@ -0,0 +1,53 @@
...
+	if input == "" {
+		return time.Time{}, nil
+	}
+
+	layouts := []string{
+		time.RFC3339,          // e.g. "2006-01-02T15:04:05Z07:00"
+		"2006-01-02",          // e.g. "2006-01-02"
+		"2006/01/02",          // e.g. "2006/01/02"
+		"2006-01-02 15:04:05", // e.g. "2006-01-02 15:04:05"
+		"01/02/2006 15:04",    // e.g. "01/02/2006 15:04"
```
omar-polo, line 34: nit: don't think the examples in here are useful :P

(also, `01/02/2006` is -unfortunately- january 2nd and not february the first.  not sure we want to keep this format since it's a huge cause of confusion between us and europe).

omar-polo (approved):
lgtm

cmd/plakar/utils/time.go
poolpOrg: indeed, me tab-tab-tab with copilot and I left the comments.

I removed the last format while at it

PR #508, opened 2025-02-16
Poolp org/adapt cli to new filters

poolpOrg:
check, ls, locate, rm now support snapshot filtering


omar-polo (approved):
nit: this is missing the flags in the SYNOPSIS; otherwise it lgtm

brmzkw (approved):

PR #521, opened 2025-02-17
fix handling of xattrs

omar-polo:
We concatenate the path with the name of the xattr, and then feed this constructed bogus path to NewExtendedAttributeReader which can't work.

Instead, keep separate the two information until we actually have to insert something in the btree.


poolpOrg (approved):
LGTM

brmzkw (approved):

PR #527, opened 2025-02-17
differentiate between resource versions for btree root wrapper and nodes

omar-polo:
Unfortunately another break, but I think we need to differentiate between the root wrapper structure and the node structures, since they're completely different structures.  This makes both recovering easier to do, but also simplifies some future work I'd like to do with sync/colouring.


poolpOrg (approved):
LGTM

mathieu-plak (approved):

PR #536, opened 2025-02-18
support pathnames relative to the scandir

poolpOrg:


snapshot/restore.go
```diff
@@ -108,6 +107,10 @@ func snapshotRestorePath(snap *Snapshot, fsc *vfs.Filesystem, exp exporter.Expor
 		}
 		defer rd.Close()
 
+		if err := exp.CreateDirectory(path.Dir(dest)); err != nil {
```
mathieu-plak, line 110: unrelated?

snapshot/restore_test.go
```diff
@@ -38,14 +39,13 @@ func TestRestore(t *testing.T) {
...
 		require.NoError(t, err)
-		fmt.Println(pathname)
 		if strings.Contains(pathname, "dummy.txt") {
 			filepath = pathname
 		}
 	}
 	require.NotEmpty(t, filepath)
 
-	err = snap.Restore(exporterInstance, fmt.Sprintf("%s/dummy.txt", exporterInstance.Root()), filepath, opts)
+	err = snap.Restore(exporterInstance, fmt.Sprintf("%s/", exporterInstance.Root()), filepath, opts)
```
mathieu-plak, line 48: unrelated?

scheduler/config.yaml
```diff
@@ -31,5 +31,8 @@ agent:
           direction: both
           peer: /tmp/foobar
 
-      cleanup:
+      retention:
```
mathieu-plak, line 34: unrelated?

snapshot/restore.go
poolpOrg: nope, problem is that as long as we restore directories we know the directories are necessarily created, that's not the case if you restore a file: it will fail because its parent directory does not exist.

mathieu-plak (approved):

snapshot/restore_test.go
poolpOrg: nope, this is supposed to be a directory not a file

omar-polo (approved):
lgtm

PR #553, opened 2025-02-21
Validate scheduler configuration

brmzkw:
To parse the configuration file, we use viper which supports yaml, toml, hcl, json, but also consul and etcd.

We use a custom mapstructure decoder to decode `BackupConfig.Check`. This way, users can provide check as:

```yaml
agent:
  tasks:
    - backup:
        check: true
```

or as:

```yaml
agent:
  tasks:
    - backup:
        check:
          enabled: true
```

The second, longer form will allow to specify more options in the future.

Also, extra unexpected keys are reported as errors (`ErrorUnused: true`).

I still have to check how to improve the error message when a task is defined without any of backup/check/restore/sync, but it can be done later.


omar-polo (approved):
don't mind using viper; lgtm

mathieu-plak (commented):
mapstructure is unmaintained, according to this https://gist.github.com/mitchellh/90029601268e59a29e64e55bab1c5bdc (link found in project description) we should switch to the maintained fork :
https://github.com/go-viper/mapstructure

brmzkw:
oops indeed, fixed

PR #559, opened 2025-02-24
do not backup a path that has the repository root as a prefix

poolpOrg:


omar-polo (approved):
not sure how this could work with a non-fs backend but OK for me, we don't want to have infinitely growing backups :P

poolpOrg:
> not sure how this could work with a non-fs backend but OK for me, we don't want to have infinitely growing backups :P

This won't work with non-fs, in practice it is a safety net to avoid accidentally backing up a repository that's within your home directory, but we can come up with something more generic if we do an adjustment to the importer, exporter and repository interfaces

mathieu-plak (approved):

PR #565, opened 2025-02-26
cleanup: Fix concurrent colouring and backup issues.

mathieu-plak:
* When we have pushed the colored state we know no one will use the blobs anymore (as they are in fact unreachable), but this leaves a window for a concurrent backup with the colouring phase that would use one of the colored blobs. To fix this when doing the sweep phase just recheck reachability of the packfile.

* While this was an heavy operation, we introduce a maintainance cache which gets updated on every run and makes the operation "is X packfile used" super fast, making the double check "free".


poolpOrg (approved):
ok with the cutoff day pushed to 30 days

omar-polo (approved):

PR #566, opened 2025-02-26
cleanup: Fix for deleted Packfiles.

mathieu-plak:
* GetSubpartForBlob() is not the best place to test if a blob belongs to a deleted packfile. Indeed, there are some rare cases (concurrent backup and colouring cleanup phase) where we might have falsely flagged a packfile for deletion, only to roll it back later at sweep time. As such, it's valid to call GetSubpartForBlob(), since we still want people to be able to acces them.

* Rather, let's do this check in the BlobExists function, so that anyone trying to push new stuff will not see it as available.

* Note that this might endup doing a bit of duplication, if we imagine a scenario where we have a concurrent backup using blob X, but we marked X for deletion, later on if we have another backup that ends up with X BlobExists() will return false so we will commit the same blob to another packfile. This is no problem (a part from disk usage), and will be adressed anyway later on by maintainance merging same blobs from different packfiles.

### DIFF
--- a/repository/repository.go
+++ b/repository/repository.go
@@ -686,16 +686,6 @@ func (r *Repository) GetBlob(Type resources.Type, mac objects.MAC) (io.ReadSeeke
 		return nil, ErrPackfileNotFound
 	}
 
-	has, err := r.HasDeletedPackfile(loc.Packfile)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


poolpOrg (approved):
OK

PR #584, opened 2025-02-28
cmd: Move diag to its own package and rename types.

mathieu-plak:
* This way we can actually use it in the RPC agent implementation. Otherwise it collides with info.


omar-polo (approved):
yay

PR #602, opened 2025-03-02
fix usage in `create` subcommand

semarie:
the `on repository` syntax is invalid and it should be `at repository`.

the `on REPOSITORY` syntax isn't accurate. it should be `at REPOSITORY`.


omar-polo (approved):
good catch, thanks!

poolpOrg (approved):
OK

Issue #600, opened 2025-03-02
options processing with "at repository" problem


semarie:
**Describe the bug**
The parsing of options (`-config`, `-cpu`, ...) is defective when using `at REPOSITORY` before the provided options (which is the documented place for putting the options).

**To Reproduce**

The following command fails:
```
$ plakar at /tmp/plakar -keyfile /tmp/plakar.key version
plakar: failed to open the repository at /tmp/plakar: open /tmp/plakar/CONFIG: no such file or directory
To specify an alternative repository, please use "plakar at <location> <command>".
```

Whereas this one is fine:
```
$ plakar -keyfile /tmp/plakar.key at /tmp/plakar version
v1.0.0-beta.1
```

**Expected behavior**

The usage format document `OPTIONS` after `at REPOSITORY`. But the code seems to not properly parse this format (it is fine with `OPTIONS` before `at REPOSITORY`).

```
$ plakar -help
Usage: plakar [at REPOSITORY] [OPTIONS] COMMAND [COMMAND_OPTIONS]...
[...]
```

**Desktop:**
 - OS: OpenBSD
 - Version -current (upcoming 7.7)

PR #603, opened 2025-03-02
fix usage string: move `OPTIONS` before `at REPOSITORY`

semarie:
the code doesn't properly parse them if having it after. so reflects the reality and provides accurate usage string.

Fixes: #600


poolpOrg (approved):
OK

PR #610, opened 2025-03-03
state: Bring back sanity check on GetBlob.

mathieu-plak:
* Since we know that no concurrent backups and maintenance should happen (as we take a repo lock), it is safe to move the check inside GetBlob so that we can return errors. Accessing a deleted blob, if there was no concurrent backups is definitely a sign of an error in our maintenance code and it should be flagged as such.

* Remove the check in BlobExists otherwise we will never hit the sanity check in GetBlob.


poolpOrg (approved):
OK

PR #624, opened 2025-03-03
more manpages fixes

semarie:
I used `mandoc -Tlint` to found and fixes some problems in man pages.

I did separated commits as some might not be not desirable (like "input text line longer than 80 bytes" or "snapshotID:path unification")


omar-polo (approved):
love the changes; thank you!

mathieu-plak (approved):
great! Thank you. And tw=80 rules. !

PR #629, opened 2025-03-04
importer/fs: Fix path cleaning.

mathieu-plak:
* Before this, this is the record we produced when doing a backup like this:

./plakar backup -quiet /tmp/powerlog/
-> record /
-> record /tmp
-> record /tmp/powerlog/

That last record, with an appended slash doesn't fare well with our VFS layer and it breaks in all kind of ways.

* Let's sanitize the path once and for all when instantiating the fs importer so that the rootDir is always a path that we can process and we can respect the VFS layer contract.

* While here remove clean from the walkdir function, it's not the right place to do this.


omar-polo (approved):
good catch!

PR #631, opened 2025-03-04
API download: add extension to filename

brmzkw:
Provide a default extension when creating an archive, if the ?name is not given or if it does not include an extension.


poolpOrg (approved):
OK

mathieu-plak (approved):

PR #636, opened 2025-03-04
trace: use hexadecimal output for lockID in trace

semarie:
when using `plakar -trace all` prefer hexadecimal output instead of raw string.


mathieu-plak (approved):
Nice catch.

poolpOrg (approved):

PR #641, opened 2025-03-05
packer: Track the goroutines using an errgroup.

mathieu-plak:
* Remove panics, deal with errors and bubble them up using the errgroup.

* In the future this really should be cancellable but for now is a fix good enough (TM)


poolpOrg (approved):
definitely OK

brmzkw (approved):

PR #648, opened 2025-03-06
packfile: Make deserialization faster.

mathieu-plak:
* Passing a pointer to a slice makes binary.Read default to reflection rather than its fast path (which is static type matching), so remove the pointer indirection we don't need it.

* This is a 60% perf gain on GetPackfile.


poolpOrg (approved):
Nice catch 🥰

omar-polo (approved):

PR #667, opened 2025-03-10
parallelize restore, almost doubles performances on my machine

poolpOrg:


snapshot/restore.go
```diff
@@ -23,104 +23,99 @@ type restoreContext struct {
...
-		snap.Event(events.DirectoryEvent(snap.Header.Identifier, pathname))
-
-		if pathname != "/" {
-			if err := exp.CreateDirectory(dest); err != nil {
-				snap.Event(events.DirectoryErrorEvent(snap.Header.Identifier, pathname, err.Error()))
-				return err
+func snapshotRestorePath(snap *Snapshot, exp exporter.Exporter, target string, opts *RestoreOptions, restoreContext *restoreContext, wg *sync.WaitGroup) func(entrypath string, e *vfs.Entry, err error) error {
+	return func(entrypath string, e *vfs.Entry, err error) error {
+		wg.Add(1)
+		go func(entrypath string, e *vfs.Entry, err error) {
```
mathieu-plak, line 29: Shouldn't this goroutine also be accounted for in the maxConcurrency semaphore? Also It's a bit unclear to me why we do need two goroutines, why can't we do all the processing in this one?

mathieu-plak: Oh okay I see your comment about deadlocking. So let me double down on this, why not spawn only one and account for that one?

omar-polo (approved):

mathieu-plak (approved):

PR #670, opened 2025-03-11
Backup checkpoints

mathieu-plak:
* Flagging as WIP as this is still a bit rough around the edges (especially regarding error reporting).

* This needs to be tested heavily obviously, but it turned out quite simple after a few iterations.

* A few stuff needs to be assessed: 

- * The fact that we now push delta state directly into the local aggregated state. Rendering (on the same machine) RebuildState virtually a no-op. This is done to avoid dealing with lookup inside multiple DeltaState (the previous flushed one) and the complexity that comes with that (been there done that). I don't think there is _any_ need for a rollback mechanism because those states are unreachable, so they'll be collected by maintenance, OR they can be reused by a backup on the same machine (since it's in its own aggregated state), as a "real time" checkpoint. Still this is enough of a big deviation from what we had before that this needs a lot of scrutiny!
- * The use of sync.Cond, I find it easier to reason about than a channel, but I'm biased by years of C, and I realise this is not idiomatic go.
- * The "end of backup" could overlap between a Flush and the final Commit (that's why I have the sync.Cond to avoid Commit actually overlapping), could we be smarter and delay a bit the Flush ? Open question.
- * The overall architecturing is a bit messy, we have stuff in Snapshot, we have stuff in Backup only structures...
- * Last but definitely not least, the error handling inside the flushDeltaState() goroutine, I'm totally open to proper way to handle those...


snapshot/packer.go
```diff
@@ -156,19 +136,10 @@ func (snap *Snapshot) GetBlob(Type resources.Type, mac [32]byte) ([]byte, error)
...
-		return snap.deltaState.BlobExists(Type, mac) || snap.repository.BlobExists(Type, mac)
-	} else {
-		return snap.repository.BlobExists(Type, mac)
-	}
+	return snap.repository.BlobExists(Type, mac)
 }
 
 func (snap *Snapshot) PutBlobIfNotExists(Type resources.Type, mac [32]byte, data []byte) error {
 	snap.Logger().Trace("snapshot", "%x: PutBlobIfNotExists(%s, %064x) len=%d", snap.Header.GetIndexShortID(), Type, mac, len(data))
-	if snap.BlobExists(Type, mac) {
```
poolpOrg, line 170: not sure why ?

poolpOrg (approved):
OK

PR #675, opened 2025-03-12
cache: Fixup iteration.

mathieu-plak:
* In GetDeltasByType() and GetDeleted() the prefix needs to include the semi colon otherwise when getting types 1 and 2 you also get types from 10 to 19 and 20 to 29 respectively.


omar-polo (approved):
ouch

omar-polo:
(actually, good catch)

PR #684, opened 2025-03-14
server: fix usage -allow-delete

semarie:
-allow-delete  will enable delete operations (and not disable them)


omar-polo (approved):
nice catch

PR #686, opened 2025-03-14
do not reuse the maxConcurrency semaphore in importerJob, not only

poolpOrg:
it is useless but it falsifies accounting and causes a deadlock


omar-polo (approved):
YUP

mathieu-plak (approved):

PR #703, opened 2025-03-18
Fix manpage for plakar config

brmzkw:


omar-polo (approved):
LGTM

can you also please run `./cmd/plakar/subcommands/sync-man.sh` to fix the markdown as well?   Thanks!

PR #708, opened 2025-03-19
Fix s3 for blackblaze and alternative providers

brmzkw:
Some S3-compatible providers, such as Backblaze, do not support the x-amz-checksum-algorithm header and return an error when it is used.

To improve compatibility, the client now computes and sends an MD5 checksum instead. This ensures broader provider support at the cost of additional resource usage.

In the future, we can make this behavior configurable if needed.


poolpOrg (approved):
OK

mathieu-plak (approved):

PR #714, opened 2025-03-19
add subcommand mount test

sayoun:
Fuse server is not daemonized, so we start the mount in a goroutine,

but the tests is not complete as we cant acces the content of the mounted repository from the tests, os.Stat returns an error everytime.

### DIFF
--- a/cmd/plakar/subcommands/mount/mount_test.go
+++ b/cmd/plakar/subcommands/mount/mount_test.go
@@ -1,0 +1,164 @@
+package mount
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/PlakarKorp/plakar/appcontext"
+	"github.com/PlakarKorp/plakar/caching"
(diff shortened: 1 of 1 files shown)


cmd/plakar/subcommands/mount/mount_test.go
```diff
@@ -0,0 +1,164 @@
...
+	require.Equal(t, "mount", subcommand.(*Mount).Name())
+
+	go func() {
+		status, err := subcommand.Execute(ctx, repo)
+		require.NoError(t, err)
+		require.Equal(t, 0, status)
+	}()
+
+	// wait for the goroutine to be executed
+	time.Sleep(1 * time.Second)
```
poolpOrg, line 141: can't it wait on a channel to be closed by the goroutine instead ?

sayoun: @poolpOrg the issue is that the goroutine will never finish, as once executed the mount command awaits until we Ctrl+C . So using a channel or a waitgroup will make the test hang indefinitely

PR #717, opened 2025-03-20
Jcastets/fix agent config

brmzkw:
Before this PR, the client was only providing the storage location and the repository passphrase to the agent. The agent was unable to open repositories that required more than just the location (access key, secret key, use tls, …).

With these changes, the client sends a RPC requests that contains:
* the name of the RPC (as before)
* the subcommand object (as before)
* the storage parameters (new)

The agent unserializes the storage parameters and use them to open the repository.

Also, to avoid code duplication, I resolve the `@` syntax in `config.GetRepository`. All the callers to `config.GetRepository` were doing the alias resolution, so moving it to `config.Repository` is just avoiding code duplication. @mathieu-plak told me he doesn't like having the resolution there, but I did not know where to put it.
If you want me to move it somewhere else, can you please suggest a place where to put the alias resolution?

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -34,7 +34,7 @@ var (
 	ErrWrongVersion   = errors.New("agent has a different version")
 )
 
-func ExecuteRPC(ctx *appcontext.AppContext, cmd subcommands.Subcommand) (int, error) {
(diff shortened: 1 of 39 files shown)


omar-polo (approved):
the encoding/decoding maybe should have been a little bit more succint but otherwise this looks good to me!

mathieu-plak (approved):

PR #720, opened 2025-03-21
since the switch from Checksum to MAC, a vfs repository can't be

poolpOrg:
shared between multiple repositories as they MAC will mismatch


omar-polo (approved):
the rubber duck likes it

ericfaurot (approved):
looks good

PR #722, opened 2025-03-21
better handling of symlinks

omar-polo:
This fixes two things actually:

* add more prefix directories to the backup so that both the user expectations and the backup navigation still works as expected
* fixes the behavior of `ls`; now it implicitly follows the symlinks inside the path.

Still a wip because Realpath() has to go away, we need an API that opens an entry following each symlinks along the path, but need to take care that Walkdir doesn't do that, otherwise we might loop indefinitely.

### DIFF
--- a/snapshot/vfs/vfs.go
+++ b/snapshot/vfs/vfs.go
@@ -1,9 +1,11 @@
 package vfs
 
 import (
+	"fmt"
 	"io"
 	"io/fs"
 	"iter"
+	"os"
 	"path"
 	"strings"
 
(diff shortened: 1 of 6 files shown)


omar-polo:
So the plan is to have `GetEntry` resolving the symlinks, but not the other functions.  There's some annoyances, like the physical vs logical path of an entry (see ls changes), but otherwise it should be transparent on other usages.

In particular, `Getdents()` doesn't have issues since it uses the physical path, so that walkdir doesn't follow symlinks except on the given argument.  (e.g. if you walk a symlink it gets resolved, otherwise it's just reported as symlink and that's it.)

snapshot/vfs/vfs.go
```diff
@@ -310,8 +311,70 @@ func (fsc *Filesystem) Pathnames() iter.Seq2[string, error] {
 	}
 }
 
-func (fsc *Filesystem) GetEntry(path string) (*Entry, error) {
-	return fsc.lookup(path)
+func (fsc *Filesystem) GetEntry(entrypath string) (*Entry, error) {
+	if !strings.HasPrefix(entrypath, "/") {
```
poolpOrg, line 315: is this necessary ?

do we ever enter this without a prefix / ?

if that's the case, shouldn't it do more than just slap a / upfront ?

omar-polo: it was in lookup, which in turns had it because the pre-btree code had something like it.  i guess it's just cargo culting from my part, and that it could be deleted.  i'll do a sweep.

poolpOrg (approved):

mathieu-plak (approved):

PR #729, opened 2025-03-24
backup: Actually save the state ID during backup.

mathieu-plak:
* Now that we insert Deltas along the process of backing up into our local state, we should also insert the State so that rebuild doesn't pick it up on a next run. This wasn't a problem (just a waste of time re-inserting same K/Vs) but is a problem for Glacier support where we want to be able to list snapshot without doing any requests at all to the repository (so no rebuild of state).


poolpOrg (approved):

ericfaurot (approved):
OK

PR #730, opened 2025-03-24
Add an objects.RandomMAC() function to generate random identifiers.

ericfaurot:


objects/objects.go
```diff
@@ -43,6 +45,19 @@ func (m *MAC) UnmarshalJSON(data []byte) error {
...
+	var r MAC
+
+	n, err := rand.Read(r[:])
+	if err != nil {
+		return MAC{}, err
+	}
+	if n != len(r) {
+		return MAC{}, io.ErrShortWrite
+	}
+	return r, nil
```
omar-polo, line 58: nit: re-reading `crypto.Read` documentation, it says that it never actually fails and always filles the given slice (but may crash on legacy Linux versions).  Now, it probably makes sense to keep the error checking, but maybe the short write check could go away.

anyway, LGTM, thank you!

omar-polo (approved):

mathieu-plak:
Thx for doing a proper version, mine was too narrow in snapshot. There are a few instances we can hunt down and replace with this helper.

mathieu-plak:
> Thx for doing a proper version, mine was too narrow in snapshot. There are a few instances we can hunt down and replace with this helper.

Actually you did kill all of them, I thought we had more. Nice work.

objects/objects.go
ericfaurot: If it cannot fail, it's better to just remove the error check completly.
It makes the code simpler.

PR #731, opened 2025-03-24
storage: change the way the backends are registered

omar-polo:
There's little point in having a Register function and then hardcode all the possible choices in New().  Instead, allow the backends to register multiple names (like "http" and "https") and do the lookup on the protocol.  Of course, "fs" is a bit special and is the fallback.


mathieu-plak (commented):
Okay but please also kill the mutex, it doesn't serve any purpose.

mathieu-plak (approved):
Could we hardcode the fallback to FS in here?

PR #747, opened 2025-03-25
Remove unreachable code intended for the "server" command

ericfaurot:
The "server" command is executed early, so all subsequent tests for that value are actually useless.


omar-polo (approved):
LGTM

When i reworked the server code to just serve the raw storage as-is, I forgot to update this as well.

thank you!

poolpOrg (approved):
LGTM

PR #748, opened 2025-03-25
add subcommand server test

sayoun:
only test one route on the server


cmd/plakar/subcommands/server/server_test.go
```diff
@@ -0,0 +1,105 @@
...
+	require.NoError(t, err, "reading deserializing configuration")
+
+	configInstance, err := storage.NewConfigurationFromBytes(version, unwrappedConfig)
+	require.NoError(t, err, "reading configuration")
+
+	// we dont test all the field from configuration
+	require.Equal(t, versioning.FromString(storage.VERSION), configInstance.Version)
+
+	// Close the goroutine by canceling the context
+	cancel()
```
omar-polo, line 104: nit: since you're already doing `defer cancel()` do we need this here as well?  Otherwise, LGTM

sayoun: you're right, removed in https://github.com/PlakarKorp/plakar/pull/748/commits/04024ab7976f9bd63ef2e729e29641eb4f95022b

mathieu-plak (approved):

PR #750, opened 2025-03-26
add subcommand sync test

sayoun:


omar-polo (approved):
lgtm; maybe some common code of GenerateRepository and GenerateSnapshot could be de-dup'ed, but can be done later if you prefer.

mathieu-plak (approved):

PR #753, opened 2025-03-27
api: serve text/html files as text/plain

omar-polo:
I couldn't help but also simplify the logic in here a bit, so please read commit-by-commit.


poolpOrg (approved):
OK

PR #758, opened 2025-03-27
env var knobs

mathieu-plak:


omar-polo (approved):
lgtm; do you want me to document these variables in plakar.1 (and in case which ones) or do we consider them an implementation detail?

mathieu-plak:
> lgtm; do you want me to document these variables in plakar.1 (and in case which ones) or do we consider them an implementation detail?

Nah we want them "hidden". People shouldn't play with this.

poolpOrg:
yes, it's just to ease the debug scenarios, people playing with undocumented variables will join the FAFO team.

PR #760, opened 2025-03-27
check jwt token before processing snapshot_path parameter

semarie:
Move the `SnapshotPathParam()` block to after JWT cookie validity checking.

It avoids:
- parsing untrusted stuff before auth
- leaking if a particular snapshot is present in the repo or not


omar-polo (approved):
completely agree, thank you!

poolpOrg (approved):

PR #761, opened 2025-03-27
caching: Switch to pebble as our cache backend.

mathieu-plak:
* Overall this gives us a nice perf boost, only regression is on smaller repository on second runs (see below), I'm still investigating that.

  (1) Korpus runs :
    (a) leveldb:
        - first run:
        real    9m0.073s

        - second run:
        real    2m15.793s

    (b) pebble:
        - first run:
        real    6m17.241s ===> -30.18%

        - second run:
        real    2m4.928s  ===> -8.14%

  (2) Linux source tree:
    (a) leveldb:
        - first run:
        real    0m48.621s

        - second run:
        real    0m6.102s

    (b) pebble:
        - first run:
        real    0m42.369s ===> -12.96%

        - second run:
        real    0m24.998s ===> +308.197%

As you can see nice win over all apart from the second run on a small-ish source. I don't think it's blocking progress for now.

* While we are here take the opportunity to factorize the code a bit. That's the first time I have what I consider a valid usage of struct Embedding.

* Warning: This bumps the cache version for obvious reasons.

github note: Markdown is bad, if you want to properly read the commit it'll be inside your pager.


poolpOrg (approved):
OK

omar-polo (approved):

PR #763, opened 2025-03-28
pebble: Disable pebble WAL logging.

mathieu-plak:


poolpOrg (approved):
oooouh I should have known, I did the same elsewhere :-)

PR #768, opened 2025-03-29
perform a constant time comparison, suggested by @semarie

### LINKED ISSUE #756: api token should be checked using constant time comparaison
**Describe the bug**
The api token should be checked using constant time comparaison.

https://github.com/PlakarKorp/plakar/blob/c7d324ec773a250a8f11a2d112c849d894456de1/api/api.go#L104-L107

The `strings.Compare()` function should be replaced by comparaison in constant time to avoid leaking information on the token.
For example, using [crypto/subtle ConstantTimeCompare()](https://pkg.go.dev/crypto/subtle#ConstantTimeCompare)

poolpOrg:
fixes #756

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1,12 +1,12 @@
 package api
 
 import (
+	"crypto/subtle"
 	"encoding/json"
 	"errors"
 	"io/fs"
 	"log"
 	"net/http"
-	"strings"
 
 	"github.com/PlakarKorp/plakar/cmd/plakar/utils"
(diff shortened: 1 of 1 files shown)


omar-polo (approved):
weird the choice to return 1/0 instead of either +1/0/-1 or a boolean, given that go has them.

anyway, love this, i should have seen it.

mathieu-plak (approved):

PR #769, opened 2025-03-29
escape ansi codes in pathnames and error

poolpOrg:


cmd/plakar/utils/utils.go
```diff
@@ -307,3 +308,15 @@ func NormalizePath(path string) (string, error) {
 
 	return normalizedPath, nil
 }
+
+var ansiEscapeSeq = regexp.MustCompile(`\x1b\[[0-9;?]*[ -/]*[@-~]`)
```
semarie, line 312: the regex is both:
- too specific : match only some of `ESC` escapes codes
- not enough specific : doesn't match some others control characters (like `BELL` or `BS` aka backspace)

it might be better to just escape all C0 controls chars (0x00 to 0x1f). `ESC` is 0x1b, so it is in the range.

see [ECMA-48](https://ecma-international.org/publications-and-standards/standards/ecma-48/), section 5.2.

semarie: and also `CSI` should be escaped as it introduces Control Sequence, see 5.4 (it is 0x9b, if I assume only 8-bits   character set to be supported).

omar-polo (commented):
i agree with @semarie ; I'd just follow what ingo@ lovely utf8.c does (see for e.g. src/bin/ls/utf8.c) and replace every non-printable character with `?` in the output.

We might use [`unicode.IsPrint`](https://pkg.go.dev/unicode#IsPrint) instead of the regex, too.  (which is not locale aware, that actually might be a problem)

also, for the logging I'd try to plug that in the logging layer itself, so that we have no fear of forgetting to escape a value, rather than on every invocation.

omar-polo:
as discussed this morning, i'm going to close this in favour of #774

PR #773, opened 2025-03-31
tests/mount: Repair darwin.

mathieu-plak:
* On macos/darwin mount needs a kext loaded, this is too much of a high entry barrier for tests. Let's disable the mount test on darwin.


omar-polo (approved):
fwiw i'm fine with this

poolpOrg (approved):

PR #780, opened 2025-04-01
entry: Make the MAC a public member.

mathieu-plak:
* The only accessor does nothing with it so no need for it to be private.

* We need it in the API so serialize it for json and omit it in msgpack to avoid recursing data structures!


omar-polo (approved):
alright, yesterday i forgot that we can use `msgpack:"-"` to avoid issues with the on-disk serialization (and thus breaking the repo format).

LGTM

poolpOrg (approved):

PR #781, opened 2025-04-01
vfs: fix vdir.ReadDir

omar-polo:
I can't read: to return all the entries, a caller may use `n <= 0`, not necessarely `n == -1`.

### DIFF
--- a/snapshot/vfs/entry.go
+++ b/snapshot/vfs/entry.go
@@ -358,13 +358,8 @@ func (vf *vdir) ReadDir(n int) (entries []fs.DirEntry, err error) {
 		}
 	}
 
+	all := n <= 0
 	for vf.iter.Next() {
-		if n == 0 {
-			break
-		}
-		if n > 0 {
-			n--
-		}
 		path, csum := vf.iter.Current()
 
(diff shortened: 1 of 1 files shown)


snapshot/vfs/entry.go
```diff
@@ -386,9 +381,15 @@ func (vf *vdir) ReadDir(n int) (entries []fs.DirEntry, err error) {
...
+
+		if !all {
+			if n--; n == 0 {
+				break
+			}
+		}
 	}
 
-	if len(entries) == 0 && n != -1 {
+	if len(entries) == 0 && !all {
```
omar-polo, line 392: actually this may just be `if len(entries) == 0` instead; just realized.

mathieu-plak (approved):
With the  !all condition removed as you noted this is Ok;

snapshot/vfs/entry.go
omar-polo: done

omar-polo:
Done! :)

omar-polo:
sorry, couldn't resist to also semplify a bit the logic :P

PR #782, opened 2025-04-01
compression: Implement a pool of lz4 readers.

mathieu-plak:
* Lz4 maintains its own buffer pool to avoid putting too much pressure on the GC. Due to the way we do reads (individual GetBlob) we instantiate lz4 readers too frequently, therefore nullifying the the benefits of the internal buffer pool. To remediate this let's pool the lz4 readers instance ourselves.

* This gives us huge wins whenever we traverse the VFS:

 (1) Maintenance:
    - before:
        real    23m55.998s
    - after:
        real    5m41.460s ===> -76%

 (2) Check:
    - before:
        real    27m1.316s
    - after:
        real    3m35.044s ===> -86%

 Other type of read heavy loads (restore/sync) should get a nice
performance boost too.

* The write (deflate) side doesn't benefit from pooling after testing, surely because they do not do buffer pooling for the write path.

### DIFF
--- a/compression/compression.go
+++ b/compression/compression.go
@@ -4,6 +4,7 @@ import (
 	"compress/gzip"
 	"fmt"
 	"io"
+	"sync"
 
 	"github.com/pierrec/lz4/v4"
 )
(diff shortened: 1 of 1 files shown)


poolpOrg (approved):

omar-polo (approved):
lovely!

PR #783, opened 2025-04-01
add subcommand agent test

sayoun:
only test one command for now

couldn't test the daemonized version because of all the os.exit(0) used in the agent code causes a panic in tests each time they are called and the test stops his execution.

### DIFF
--- a/cmd/plakar/subcommands/agent/agent_test.go
+++ b/cmd/plakar/subcommands/agent/agent_test.go
@@ -111,7 +111,7 @@ func TestCmdAgentForegroundInit(t *testing.T) {
 		}
 	}()
 
-	client, err := agent.NewClient(filepath.Join(ctx.CacheDir, "agent.sock"))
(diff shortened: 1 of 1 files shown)


omar-polo (approved):
ops, apologize.  i've missed this and now i've also broke it :(

tl;dr after a repository and snapshot change i did a sweep to fix the tests after the fallout and in doing so i've changed a little bit how GenerateSnapshot works.

I'll propose a fix to this branch as well.

Thank you!

omar-polo:
the restore test is sometimes failing due to a timing issue with the log flush; fwiw i've reworked the test a little bit here: https://github.com/PlakarKorp/plakar/pull/830

sayoun:
sorry I was off for a couple of days, thanks for the restore test fix PR and this one also @omar-polo  !

poolpOrg (approved):

PR #785, opened 2025-04-02
api: allow searching files with several mimes

brmzkw:
Also, the command plakar diag <snapshot> <mimes> now accepts a commas separated string.

### DIFF
--- a/snapshot/search.go
+++ b/snapshot/search.go
@@ -12,7 +12,7 @@ type SearchOpts struct {
 	// file filters
 	Recursive  bool
 	Prefix     string // prefix directory
-	Mime       string
+	Mimes      []string
 	NameFilter string
 
 	// pagination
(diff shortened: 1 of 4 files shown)


snapshot/search.go
```diff
@@ -51,65 +53,67 @@ func visitmimes(snap *Snapshot, opts *SearchOpts) (iter.Seq2[*vfs.Entry, error],
...
+				prefix += "/"
 			}
 
-			entry, err := fsp.ResolveEntry(mac)
+			// XXX: to further optimize this routine we'd need a ScanGlob
+			// method.  Basically what we're trying to do here is to
+			// resolve globs like "application/*/prefix/*"
+			it, err := idx.ScanFrom(prefix)
 			if err != nil {
 				yield(nil, err)
```
omar-polo, line 68: nit: here we're missing a return afterwards, just in case the calling code don't `break` after the error.

omar-polo (approved):
i've left a comment but otherwise it looks good to me.  I'd still like to optimize these kinds of glob queries, but for now it's fine.

another nit is that we might pre-compute the `SplitN` on the mimes array to avoid doing it in the loop, but it may not be *that* costly.

snapshot/search.go
brmzkw: fixed

poolpOrg (approved):
OK will improve in tree

PR #795, opened 2025-04-04
Fix plakar backup usage

brmzkw:
The syntax `plakar backup s3://...` is not valid anymore. To backup a remote, you should use the @ syntax and configure the remote in ~/.config/plakar/plakar.yml.


omar-polo (approved):
yeah, but the manpage has a similar issues, it only talks about a "directory" when it could have something different.  if you don't mind i'll push a fix for it as well in this branch.

poolpOrg (approved):

PR #798, opened 2025-04-05
Poolp org/storage returns size

poolpOrg:


repository/repository.go
```diff
@@ -44,6 +44,9 @@ type Repository struct {
 	configuration storage.Configuration
 
 	appContext *appcontext.AppContext
+
+	wBytes int64
+	rBytes int64
```
mathieu-plak, line 49: Those needs to be atomics they can be incremented from goroutine context.

mathieu-plak (changes requested):

repository/repository.go
poolpOrg: fixed !

mathieu-plak (approved):

omar-polo (approved):

PR #802, opened 2025-04-06
add a Size function to interface

poolpOrg:


omar-polo (approved):
very nice!

mathieu-plak (approved):
Could use a warning about the cost at the interface level. Otherwise LGTM

PR #806, opened 2025-04-07
add a bypass for critical checks

poolpOrg:


omar-polo:
actually, `shouldCheckUpdate` in `utils.go` is probably the right place where to put this IMHO, right before the cookie check.

mathieu-plak (approved):

PR #808, opened 2025-04-07
dirty flag should only be reset if dirty

poolpOrg:


repository/repository.go
```diff
@@ -253,6 +255,7 @@ func (r *Repository) Store() storage.Store {
 }
 
 func (r *Repository) StorageSize() int64 {
+	fmt.Println("StorageSize()", r.storageSize, r.storageSizeDirty)
 	if r.storageSizeDirty {
```
omar-polo, line 259: minus this OK; good catch, i missed this during review, sorry.

omar-polo (approved):

repository/repository.go
poolpOrg: removed

PR #811, opened 2025-04-07
backup: Provide a check cache if check is specified.

mathieu-plak:
* When backing up we allow the user to ask for a check of the just created snapshot. Our check code now uses a cache, so let's specify one here before running it otherwise it crashes.

* We could also make the check code work without cache, but there are benefits (speed benefits) to running with a cache, even for a single snap check.

* Crash reported by niluje@


omar-polo (approved):

poolpOrg (approved):
duh

PR #820, opened 2025-04-10
api: don't return chunks in entries; add endpoint for them

### LINKED ISSUE #812: Do not return chunks in the object
The endpoint `http://localhost:30112/api/snapshot/vfs/children/<snapshot>:<path>?offset=0&limit=50` returns the list of all children for a given path.

For a folder containing very large files, the list of chunks is returned and the response can be huge.

We should instead move the chunks to a separate endpoint, and have pagination.

omar-polo:
Fixes #812


poolpOrg (approved):
OK

mathieu-plak (approved):
Yup! json slow slow! Bad json!

PR #822, opened 2025-04-10
unbreak tests and rework Generate{Repository,Snapshot} a bit while here

omar-polo:


mathieu-plak (approved):
Thx thx thx for fixing my mess.... I owe you a beer or two.

poolpOrg (approved):

PR #829, opened 2025-04-11
PoolpOrg/parallelize-chunk-macsum

poolpOrg:


snapshot/backup.go
```diff
@@ -435,18 +444,33 @@ func (snap *Builder) chunkify(imp importer.Importer, record *importer.ScanRecord
...
+			semaphore <- struct{}{}
+			wg.Add(1)
+
+			go func(idx int, cdcCopy []byte) {
+				defer wg.Done()
+				defer func() { <-semaphore }()
+
+				if err := processChunk(idx, cdcCopy); err != nil {
+					//return nil, err
+				}
```
omar-polo, line 465: not a huge fan of this to be honest :D

we need to have a way to handle the PutBlob failure.

snapshot/backup.go
```diff
@@ -358,75 +357,85 @@ func (snap *Builder) chunkify(imp importer.Importer, record *importer.ScanRecord
...
 		}
 	} else {
+		semaphoreSize := record.FileInfo.Size() / int64(snap.repository.Configuration().Chunking.MinSize)
+		if semaphoreSize < 1 {
+			semaphoreSize = 1
+		}
+		if semaphoreSize > int64(backupCtx.maxConcurrency) {
+			semaphoreSize = int64(backupCtx.maxConcurrency)
+		}
+		semaphore := make(chan struct{}, 42)
```
omar-polo, line 429: the answer to life, the universe and everything, but maybe we should use `semaphoreSize` :P

snapshot/backup.go
poolpOrg: I was just making sure people reviewed my diffs :-p

poolpOrg: fixed by switching to an errgroup

poolpOrg:
TO BE UPDATED AND FIXED

PR #835, opened 2025-04-14
state: Fix miantenance states.

mathieu-plak:
* CommitTransaction pushes locally the deltaState ID because it expects it to be insert along the process. Make it true for DeleteRessource too.


omar-polo (approved):
obvious ok from me :-)

PR #839, opened 2025-04-16
cmd: Disallow usage of at for agent cmd.

### LINKED ISSUE #817: `plakar at <xx> agent` should fail
The command `plakar agent` doesn't operate on a repository. We should add an error if `at xxx` is given.

mathieu-plak:
Fixes #817


omar-polo (approved):
yay for more special casing :P

PR #842, opened 2025-04-17
add tests for utils package in subcommands

sayoun:


cmd/plakar/utils/locate_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	repo, snap := generateSnapshot(t, bufOut, bufErr)
+	defer snap.Close()
+
+	found, err := LocateSnapshotByPrefix(repo, hex.EncodeToString(snap.Header.GetIndexShortID()))
+	require.NoError(t, err)
+	require.Equal(t, found, snap.Header.Identifier)
+
+	_, err = LocateSnapshotByPrefix(repo, hex.EncodeToString([]byte{0x00}))
+	require.EqualError(t, err, "no snapshot has prefix: 00")
```
omar-polo, line 94: i fear that this might break from time to time if we're lucky enough to generate a snapshot that starts with 0x00.  Since we generate only one snapshot, we could do something like `snap.Header.Identifier[0] + 1` to get back a byte that's for sure not a prefix for anything.  (that sum should implicitly wrap around for 0xff to 0x00).

omar-polo (approved):
spotted a nit but overall looks good to me, thank you!

cmd/plakar/utils/locate_test.go
sayoun: good catch @omar-polo , updated in https://github.com/PlakarKorp/plakar/pull/842/commits/59ca7a82608ced3b5f97fe78583f637c3cf65dd6

omar-polo: thank you!

poolpOrg (approved):

PR #850, opened 2025-04-18
api: serve all text/* as text/plain

omar-polo:
niluje@ spotted that text/xml was still interpreted; so let's use the big hammer and serve all text/* stuff as text/plain to stay on the safe side.

The intent is to always show the content of the files in the UI, so we have to prevent the browser from trying to be 'smart'.


omar-polo:
i hereby approve @brmzkw commit!

brmzkw (approved):

poolpOrg (approved):

PR #857, opened 2025-04-22
cli: Rework the CLI/RPC infrastructure.

mathieu-plak:
* Register a factory rather than a parsing method for each CLI subcommands and use that in RPC to remove the big switch case and turn it into a lookup.

* This is not perfect yet (the fact that we pass the matched name down to RPC level for example is a pain point) but it's a really good step in the right direction (the diffstat speaks for itself)


poolpOrg:
lovely !

poolpOrg (approved):

PR #859, opened 2025-04-22
add tests for ptar command

sayoun:


poolpOrg (approved):

cmd/plakar/subcommands/ptar/ptar_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func init() {
+	os.Setenv("TZ", "UTC")
+}
+
+func TestExecuteCmdPtarDefault(t *testing.T) {
+	bufOut := bytes.NewBuffer(nil)
+	bufErr := bytes.NewBuffer(nil)
```
omar-polo, line 20: do we need these?  we never look at the output in this test, nor in the next one.

testing/snapshot.go
```diff
@@ -102,6 +122,9 @@ func GenerateSnapshot(t *testing.T, repo *repository.Repository, files []MockFil
 		f(o)
 	}
 
+	tmpBackupDir := GenerateFiles(t, files)
+	repo.AppContext().CWD = tmpBackupDir
```
omar-polo, line 126: i get the point of GenerateFiles, you need it for ptar, and that's ok for now*, but why plugging it in here as well?  We're not going to look at the `tmpBackupDir` anyway since we use the MockImporter and not the fs one.  Am I missing something?

omar-polo (changes requested):
two nits but otherwise i like it, thank you!

cmd/plakar/subcommands/ptar/ptar_test.go
sayoun: you're right, removed in https://github.com/PlakarKorp/plakar/pull/859/commits/2074f3fa6694cb5f431b9af59ee0be46fd434fa4

testing/snapshot.go
sayoun: it's not needed, it was an error after rebasing with main, I removed it in https://github.com/PlakarKorp/plakar/pull/859/commits/af072c630b6580811ae3926614596134d120c9c7

cmd/plakar/subcommands/ptar/ptar_test.go
omar-polo: thank you!

testing/snapshot.go
omar-polo: thank you!

omar-polo (approved):

PR #866, opened 2025-04-24
make plakar installable from the top level directory as well

omar-polo:
This moves the bulk of the code of the command line in a separate package, so that we can reuse it in a new top-level `main.go` file which in turns makes the installation easier to do without moving too many things.


poolpOrg (approved):
YES PLEASE !

mathieu-plak (approved):
LGTM

PR #870, opened 2025-04-24
cmd/check: Fix snapshot parsing.

mathieu-plak:
* seems like you can't do an omelet without breaking some eggs ....


poolpOrg (approved):
I don't like omelets, OK

PR #872, opened 2025-04-24
cmd: Remove the RPC interface.

mathieu-plak:
* The RPC interface relied on the Name() method to select which commands could get sent over the wire. Retire the Name method and the whole interface so that we can use Subcommand everywhere including over the wire.

* In place use the existing Flags mechanism to tell the agent which command can be used over the wire or not.

* This needs a bit of tweaking to the Lookup method, as we rely on it to set the flags on the object, meaning we don't return a factory anymore but a created (and updated) object. This also means no one should use the real objects without a Lookup() before.


agent/agent.go
```diff
@@ -96,7 +92,11 @@ func (c *Client) handshake() error {
 	return nil
 }
 
-func (c *Client) SendCommand(ctx *appcontext.AppContext, name []string, cmd subcommands.RPC, storeConfig map[string]string) (int, error) {
+func (c *Client) SendCommand(ctx *appcontext.AppContext, name []string, cmd subcommands.Subcommand, storeConfig map[string]string) (int, error) {
+	if cmd.GetFlags()&subcommands.AgentSupport == 0 {
+		return 1, fmt.Errorf("Command %v doesn't support execution through agent\n", strings.Join(name, " "))
```
omar-polo, line 97: nit: no newline in error messages please :P

omar-polo (approved):
noice

agent/agent.go
mathieu-plak: oh that wasn't supposed to be there. Nice catch.

omar-polo:
apologise for introducing the conflict :/

mathieu-plak:
> apologise for introducing the conflict :/

Np! It's fixed now.

poolpOrg (approved):

PR #873, opened 2025-04-24
{storage,importer,exporter}: unify location parsing

omar-polo:
so far i've only tested this with the fs importer, but i'm quite sure it'll work just fine for exporters as well.

For example, these three invocation are now equivalent:

```
$ pwd
/home/op/w/plakar/
$ plakar backup fs:///home/op/w/plakar/btree
[...]
$ plakar backup fs:/home/op/w/plakar/btree
[...]
$ plakar backup fs://btree
[...]
$ plakar backup fs:btree
[...]
$ plakar backup btree
[...]
```


location/location.go
```diff
@@ -0,0 +1,76 @@
+package location
+
+import (
+	"slices"
+	"strings"
+	"sync"
+)
+
+type Location[T any] struct {
+	mtx      sync.Mutex
```
mathieu-plak, line 10: We don't really need this but it doesn't hurt I guess.

mathieu-plak (approved):
love it! Finally we got around to making this generic.

poolpOrg (approved):

location/location.go
omar-polo: it's to be future-proof for when we'll have the plugins; also, it was already there :-)

PR #882, opened 2025-04-28
backup: Remove a erroneous PutBlob(OBJECT)

mathieu-plak:
* Discussed with gilles@ and op@.

* This would lead to corruption in case of an aborted backup then doing it again. Because we would fetch the RT_OBJECT out of the VFS cache but then the first thing was putting it in the repository, therefore bypassing chunkify-cation of the object.


omar-polo (approved):

poolpOrg (approved):
needs to get in fast to be tested more heavily

PR #887, opened 2025-04-29
properly pass down the context and fix NewAppContextFrom

omar-polo:
While here, also get rid of a few new contexts that are not needed: only call NewAppContextFrom when we actually need a "checkpointing" cancellation context, e.g. a network connection, or when we have to set a different secret.

This will be helpful in handling interruptions in the agent.

### DIFF
--- a/cmd/plakar/subcommands/agent/agent.go
+++ b/cmd/plakar/subcommands/agent/agent.go
@@ -17,7 +17,6 @@
 package agent
 
 import (
-	"context"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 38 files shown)


cmd/plakar/subcommands/agent/agent.go
```diff
@@ -294,7 +287,7 @@ func (cmd *Agent) ListenAndServe(ctx *appcontext.AppContext) error {
 			read := func(v interface{}) (interface{}, error) {
 				if err := decoder.Decode(v); err != nil {
 					if isDisconnectError(err) {
-						cancel()
+						clientContext.Close()
```
poolpOrg, line 290: the Close() is already deferred line 258

poolpOrg (approved):
OK

mathieu-plak (approved):
Joint work so yeah, approved :D.

cmd/plakar/subcommands/agent/agent.go
omar-polo: yeah but we still need to forcefully shut down the connection.  calling close twice is fine, only the first will really cancel the context.

omar-polo: because we are in a different routine here that's defferred below.  it's a bit of a mess.

PR #888, opened 2025-04-29
importer/fs: simplify walkdir_walker

omar-polo:
reduce the number of goroutines (and headaches reading this code); split off the realpath thingy into its own function.


omar-polo:
couldn't resist doing a few more simplifications; although it's probably better to review commit-by-commit.

mathieu-plak (approved):

poolpOrg (approved):
BEAUTIFUL ARTWORK

PR #899, opened 2025-04-30
do not forget to set randomly generated id to the cache structure,

poolpOrg:
otherwise removal will fail on bogus constructed path


mathieu-plak (approved):

omar-polo (approved):
nice catch

PR #901, opened 2025-04-30
cmd/backup: Provide the snapshot id created by the backup.

mathieu-plak:
* For now, as we still tap into the CMD for agent's scheduler, we need to access the snapshot id created out of the backup. To do this, as things usually go in computer, provide another layer of indirection to fix this. Introduce a Do function that returns the needed information, and wraps that in the CMD Execute() func. The scheduler can directly use the Do func.

* While here remove the snapshotID from AppContext, this can't work like this.


omar-polo (approved):
love it; it's also a very tiny step towards cleaning this usage of cmd in the scheduler, so YAY FOR ME

poolpOrg (approved):

PR #905, opened 2025-05-01
backends/s3: use a SectionReader for GetPackfileBlob

omar-polo:
avoids to read all the data in memory; fixes a "high security issue" reported by CodeQL.


omar-polo:
p.s. i still think that the backend interface should return a ReadCloser, will try to change it after release.

mathieu-plak (approved):
What security issue ? xD. It reads way better though

omar-polo:
@mathieu-plak codeql was complaining about a user-controllable allocation

PR #917, opened 2025-05-03
importer/fs: Use the correct gid table for memoization.

mathieu-plak:
* Spotted during a perf regression tests, the lookup function kept showing in the flamegraphs, that's because we used the GID on the LookupId() function, which meant it failed all the time bypassing memoization.

* This whole gid branch was wrong anyway, using the wrong map too.


omar-polo (approved):
ops, good catch!

poolpOrg (approved):

PR #919, opened 2025-05-03
continue the check refactoring

omar-polo:
This tidies the check code a bit more.  It also includes #918 as a side-effect of the changes, depending on what we decide to do for the release this might wait a bit, or i can rebase it and fix the trivial conflict when #918 will get in.


omar-polo:
p.s. i'd suggest reading this with whitespaces disabled; it's easier to see the actual changes.

poolpOrg (approved):
me ok

PR #927, opened 2025-05-04
we know the size of the file we want to restore

poolpOrg:
the minio client will allocate excessive amounts of memory if the size of a file is unknown at push, ... however in Restore, we always know the size.


poolpOrg:
fixes s3 client on OpenBSD

omar-polo (approved):

PR #928, opened 2025-05-04
return copy of config

poolpOrg:


poolpOrg:
forced merged as it is much needed right away, will discuss if there's a better fix when everyone's back

PR #932, opened 2025-05-06
Poolp org/fix login

poolpOrg:


ericfaurot (approved):
Will need factorizing and better error handling later

PR #936, opened 2025-05-06
Add `stdio` and `ftp` importer and exporter tests

sayoun:


connectors/stdio/importer/stdio_test.go
```diff
@@ -0,0 +1,66 @@
...
+	paths := []string{}
+	for record := range scanChan {
+		require.Nil(t, record.Error)
+		paths = append(paths, record.Record.Pathname)
+
+		if record.Record.FileInfo.Mode().IsRegular() {
+			content, err := io.ReadAll(record.Record.Reader)
+			require.NoError(t, err)
+			require.Equal(t, content, []byte("test importer stdin"))
+			record.Record.Reader.Close()
```
omar-polo: nitpick: `record.Reader.Close` has to be always called.  not very important in this case since we know it's a `LazyReader`, but still it's better to have the right pattern in place.

connectors/ftp/exporter/ftp_test.go
```diff
@@ -0,0 +1,105 @@
...
+		"file1.txt": "content1",
+		"file2.txt": "content2",
+	}
+
+	for name, content := range testFiles {
+		path := filepath.Join(tmpDir, name)
+		if err := os.WriteFile(path, []byte(content), 0644); err != nil {
+			t.Fatalf("Failed to create test file %s: %v", name, err)
+		}
+	}
```
omar-polo: I'm not sure I get why we have to physically create the files on disk

connectors/stdio/importer/stdio_test.go
sayoun: @omar-polo you're right, fixed in https://github.com/PlakarKorp/plakar/pull/936/commits/e770c5e77d0cafd66ce4eb1d07d68edca1fb3f83

connectors/ftp/exporter/ftp_test.go
sayoun: @omar-polo you're right, fixed in https://github.com/PlakarKorp/plakar/pull/936/commits/e770c5e77d0cafd66ce4eb1d07d68edca1fb3f83

PR #947, opened 2025-05-07
report commands executed through the CLI

poolpOrg:


poolpOrg:
This was worked on and okayed by @ericfaurot during a call, we'll commit as we need it in, it will be reworked and reviewed by another @PlakarKorp/developers when they're back from vacation.

PR #962, opened 2025-05-09
fix backup hang

### LINKED ISSUE #961: backup stalls on invalid import source
$plakar backup /nonexistent

would hang before be502b78d58, the fix in there needs to be revisited.

mathieu-plak:
Fixes #961.


poolpOrg (approved):
indeed, I misanalysed the root issue

PR #979, opened 2025-05-10
fix a 500 in local API when no auth token is available

poolpOrg:


poolpOrg:
reviewed with @brmzkw

PR #982, opened 2025-05-10
s3: locks do not have buckets

brmzkw:


poolpOrg:
debugged together, ok

PR #990, opened 2025-05-12
s3 importer: fix directories creation

brmzkw:
Records created with importer.NewScanRecord should not contain a trailing slash in the path, otherwise the backup layer fails to retrieve the parent directory.


omar-polo (approved):

mathieu-plak (approved):
S3 importer needs a bit more love. But yeah okay.

PR #1010, opened 2025-05-13
Use a common execution path for all commands with a repository.

ericfaurot:
Commands with the `BeforeRepositoryWithStorage` flag have their repository set up differently, but use the same exectution path as regular commands.


omar-polo (approved):
yay, lovely!

poolpOrg (approved):
stamped

PR #1011, opened 2025-05-13
make the stdin importer work in agent mode

omar-polo:
This augemnts the agent protocol so that the agent can request chunks from stdin from the client.  The AppContext then gains a new Stdin reader that transparently proxies the request to the client.

In the non-agent case, everything remains the same, we read directly from os.Stdin.

The tricky part is making sure that the EOF error is correctly propagated, but after the serialization we've lost the types, hence the not-so-pretty "EOF" string match.


poolpOrg (approved):

omar-polo:
closing in favour of https://github.com/PlakarKorp/plakar/pull/1100

PR #1014, opened 2025-05-13
entrypoint: delay the signal trapping after subcommand.Parse

omar-polo:
This way we can still ctrl-c our way out of the "repository passphrase:" prompt.


poolpOrg (approved):

ericfaurot (approved):
ok

PR #1025, opened 2025-05-14
s3: do not store the inode

brmzkw:
The functions Equal and EqualIgnoreSize of objects/fileinfo.go consider two fileinfo structs as different if the inode number is different.

Since the importer generates an inode per file, in the case a second import has one more file in the middle of the import, all the subsequent files will have a different inode and won't be considered as being cached.

As an alternative, we could also stop comparing the inodes in Equal and EqualIgnoreSize.


poolpOrg (approved):

omar-polo (approved):
lgtm

PR #1028, opened 2025-05-14
backup/flusher: Exit the flusher on context cancel

mathieu-plak:


omar-polo (approved):
i just found it funny that we had this yesterday and then we dropped it :D

poolpOrg (approved):
OK

PR #1031, opened 2025-05-15
s3 importer: improve listing performance

brmzkw:
Switch to using ListObjects with Recursive: true, which returns up to 1000 objects per API call. Previously, we implemented recursion manually by listing one directory at a time, which resulted in a large number of slow calls—especially when directories contained few entries.

Now, for each returned file, we generate records for all of its parent directories. This is safe because the backup layer ignores duplicated entries.

Performance gains are substantial: importing a bucket with ~55,000 files and ~1,200 directories (mostly already present in the repository) now completes in ~12 seconds, down from ~13 minutes.


poolpOrg:
oddly enough I had this changed done already for other reasons, so I'm generally ok with the direction, too fried to test or review today though.

omar-polo (approved):

poolpOrg (approved):

PR #1039, opened 2025-05-16
btreescan: use config + all importers.

mathieu-plak:
As a side effect this will make it easier to test (let's us swap the importer).


omar-polo (approved):
lovely, thanks!

poolpOrg (approved):

PR #1040, opened 2025-05-16
agent: tweak the command logging

omar-polo:


mathieu-plak (approved):
Yes on the diff. 
More general question, I'm wondering if logging location is a good idea? It _might_ contain credentials no? We do that in a few places (sync comes first to mind)

poolpOrg (approved):

Issue #922, opened 2025-05-03
stdin importer does not work in agent mode


poolpOrg:
stdin importer does not work in agent mode or at least it does not work when the repo is encrypted, to be investigated

poolpOrg:
this has been worked on by @omar-polo on #1011  and will be merged post release

poolpOrg:
Still does not work

```
$ cat /etc/passwd | ./plakar backup stdin:/tmp/foobar.txt
fa5dcc3f: KO ✘ /tmp/foobar.txt: read /dev/stdin: file already closed
fa5dcc3f: OK ✓ /tmp
fa5dcc3f: OK ✓ /
info: backup: created unsigned snapshot fa5dcc3f of size 0 B in 1.484583ms (wrote 546 kB)

$ cat /etc/passwd | ./plakar -no-agent backup stdin:/tmp/foobar.txt
e2bbb393: OK ✓ /tmp/foobar.txt
e2bbb393: OK ✓ /tmp
e2bbb393: OK ✓ /
info: backup: created unsigned snapshot e2bbb393 of size 9.1 kB in 32.124209ms (wrote 516 kB)
$ 
```

omar-polo:
fixed in https://github.com/PlakarKorp/plakar/pull/1100

PR #1041, opened 2025-05-16
try to use the tty for the passphrase prompts

omar-polo:
This is the second part needed for #922.  It means that it'll be possible to run the following and still being able to prompt the user for a password, assuming a tty is available:

    $ pgdump | plakar backup stdin://tmp/foo
    repository passphrase: ***
    [...]


poolpOrg (approved):
yup

ericfaurot (approved):
ok

PR #1042, opened 2025-05-16
agent: Disallow agent start with PLAKAR_AGENTLESS.

mathieu-plak:
* This can lead to weird situation like having two agent processes, or failing to send commands (if you unset it after etc).

* It makes no sense anyway so prevent it.

### DIFF
--- a/subcommands/agent/agent.go
+++ b/subcommands/agent/agent.go
@@ -93,6 +93,11 @@ func (cmd *Agent) Parse(ctx *appcontext.AppContext, args []string) error {
 	var opt_foreground bool
 	var opt_logfile string
 
+	_, envAgentLess := os.LookupEnv("PLAKAR_AGENTLESS")
+	if envAgentLess {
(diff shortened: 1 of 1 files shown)


poolpOrg:
shouldnt you also make sure we don't have `-no-agent` ?

```
$ plakar -no-agent agent
agent started with pid=71603
$ 
```

mathieu-plak:
> shouldnt you also make sure we don't have `-no-agent` ?
> 
> ```
> $ plakar -no-agent agent
> agent started with pid=71603
> $ 
> ```

I'm on the fence. Doing that means introducing yet another Cmd Flag. the env var is a silent killer, "-no agent agent" is pretty in your face.

Happy to do it if you think it's worth it.

poolpOrg:
I actually think it's legit:

When running `plakar -no-agent agent` you want to start the agent without going through the agent, so it's semantically correct, no need to introduce a special case.

poolpOrg (approved):

omar-polo (approved):
lgtm

PR #1073, opened 2025-05-20
maintenance: Remove locks now that we are out of release.

mathieu-plak:
This is part2 ....


poolpOrg:
We're way past timeout on this one:

backups are not exclusive locked protected, maintenance still is but requires a lot more R&D than we initially thought as our plan was flawed.

Closing, we'll revisit when we've made progress.

PR #1075, opened 2025-05-22
Introduce a plakar AppContext.

mathieu-plak:
* Changes all dependency to point to plakar's AppContext.

* For now plakar's AppContext is just embedding kloset's KContext. The plan is to gradually push stuff out of KContext and into plakar until we can turn KContext a simple context.Context

* This also adapts to the new connectors interface (swapping KContext for a raw context.Context)

* Goes hand in hand with https://github.com/PlakarKorp/kloset/pull/4


omar-polo (approved):
my eyes bleed now

ericfaurot (approved):
ok

PR #1084, opened 2025-05-28
stdio importer should have a reader for file but not dir

sayoun:


omar-polo (approved):
ooops, my fault.  Good catch, thank you!

PR #1085, opened 2025-05-29
appcontext: move the secret in here

omar-polo:
kloset now takes the secrets in the Repository constructors instead of implicitly thru the kcontext.  To minimize the diff, move it in the appcontext for now.  Its lifetime may be shrunk in a follow-up.

kloset pr: https://github.com/PlakarKorp/kloset/pull/8


mathieu-plak (approved):

ericfaurot (approved):
ok

PR #1091, opened 2025-05-30
ui: use path instead of filepath

omar-polo:
inside an embed filesystem we always have UNIX-style paths.  Fixes `plakar ui` on windows


omar-polo:
don't hold your breath, this is not the only thing broken on windows.  it's just one of the easiest to backport from my windows branch.

brmzkw (approved):

PR #1092, opened 2025-05-30
importers: pass ImporterOpts

omar-polo:
Needed after the kloset change:
https://github.com/PlakarKorp/kloset/pull/11

To minimize the diff add an helper method AppContext.ImporterOpts.  In the future we might want to tweak some parameters, but for now we're fine with a simple copy.


poolpOrg:
borken due to conflict

mathieu-plak (approved):

PR #1098, opened 2025-06-05
FEAT: New GRPC_exporter struct

Peralban:
This pull request introduces a gRPC-based exporter implementation for the snapshot/exporter, this will be used to communicate with all the plugins.

### New gRPC Exporter Implementation:
- Added GrpcExporter in snapshot/exporter/grpc_exporter.go, which implements methods like CreateDirectory, SetPermissions, StoreFile, and Close using gRPC client calls. This class facilitates communication with a gRPC server for snapshot exporting.


poolpOrg (approved):
OK

connectors/grpc/exporter/grpc_exporter.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type GrpcExporter struct {
+	GrpcClient grpc_exporter.ExporterClient
+}
+
+func (g *GrpcExporter) Close() error {
+	_, err := g.GrpcClient.Close(context.Background(), &grpc_exporter.CloseRequest{})
```
mathieu-plak: You can't use context.Background here. You might be in agent with a conn dedicated context.

connectors/grpc/exporter/grpc_exporter.go
```diff
@@ -0,0 +1,87 @@
...
+	for {
+		n, err := fp.Read(buf)
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return err
+		}
+		
+		if err := stream.Send(&grpc_exporter.StoreFileRequest{
```
mathieu-plak, line 86: A bit confused by the protocol here. We send N packets of type StoreFileREquest, with N time the pathname / the file size but a different chunk of the file?

connectors/grpc/exporter/grpc_exporter.go
Peralban: Do I need to use the ContextApp ?

mathieu-plak: No you need to use the one given to you by the constructor. You need a NewGRPCImporter implements the ExporterFn type (https://github.com/PlakarKorp/kloset/blob/main/snapshot/exporter/exporter.go#L23).

Otherwise plakar will never be able to instantiate your exporter.

poolpOrg: oooouh good catch

mathieu-plak (approved):

PR #1099, opened 2025-06-05
FEAT: New GRPC_storage struct

Peralban:
### New gRPC Storage Implementation
* Added `GrpcStorage` in `storage/grpc_storage.go`, providing a gRPC-based storage backend with methods for creating, opening, and managing states, packfiles, and locks. That new struct will be used for the execution of the plugins in kloset.


connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,323 @@
...
+	"github.com/PlakarKorp/kloset/storage"
+
+	grpc_storage "github.com/PlakarKorp/plakar/connectors/grpc/storage/pkg"
+)
+
+type GrpcStorage struct {
+	GrpcClient grpc_storage.StoreClient
+}
+
+const bufferSize = 4096
```
poolpOrg: 16 * 1024 ?

connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,323 @@
...
+func (s *GrpcStorage) Open(ctx context.Context) ([]byte, error) {
+	resp, err := s.GrpcClient.Open(ctx, &grpc_storage.OpenRequest{})
+	if err != nil {
+		return nil, err
+	}
+	return resp.Config, nil
+}
+
+func (s *GrpcStorage) Location() string {
+	resp, err := s.GrpcClient.GetLocation(context.Background(), &grpc_storage.GetLocationRequest{})
```
mathieu-plak: Here and below you don't want to use Background. You might be running inside the agent, you need to use the provided context at all times.

connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,323 @@
...
+		return ""
+	}
+	return resp.Location
+}
+
+func (s *GrpcStorage) Mode() storage.Mode {
+	resp, err := s.GrpcClient.GetMode(context.Background(), &grpc_storage.GetModeRequest{})
+	if err != nil {
+		return storage.Mode(0)
+	}
```
mathieu-plak, line 73: It's a bit unfortunate that we can't signal an RPC error in those non failing interfaces. Curious to hear thoughts from others here, but we are basically silencing an error here.

connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,323 @@
...
+
+func (s *GrpcStorage) Size() int64 {
+	resp, err := s.GrpcClient.GetSize(context.Background(), &grpc_storage.GetSizeRequest{})
+	if err != nil {
+		return -1
+	}
+	return resp.Size
+}
+
+//========================
```
mathieu-plak: Stray comment line.

connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,323 @@
...
+
+type chunkSenderFunc func(chunk []byte) error
+
+func sendChunks(rd io.Reader, send chunkSenderFunc) (int64, error) {
+	buffer := make([]byte, bufferSize)
```
mathieu-plak, line 86: I know the magic spewing computer told you to make a constant but actually it's a bad choice. It's not re-used anywhere... just use it here and if it's derived from experiments / a particular reason mention it in a commit.

connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,323 @@
...
+			}
+			return nil, fmt.Errorf("receive chunk: %w", err)
+		}
+		buf.Write(chunk)
+	}
+	return io.NopCloser(&buf), nil
+}
+
+func toGrpcMAC(mac objects.MAC) *grpc_storage.MAC {
+	return &grpc_storage.MAC{Value: mac[:]}
```
mathieu-plak, line 155: gah this is insane. Good thing we have an SDK...

connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,323 @@
...
+		if err != nil {
+			return totalBytes, fmt.Errorf("failed to read: %w", err)
+		}
+	}
+	return totalBytes, nil
+}
+
+type chunkReceiverFunc func() ([]byte, error)
+
+func receiveChunks(recv chunkReceiverFunc) (io.Reader, error) {
```
mathieu-plak: Wouldn't this mean that we hold everything in memory? GetState might be a problem then.... but then again apart from writing the file on disk I don't see many options.

But it was an hard requirement for us to never fully load state files in memory and this feels like a stepback.

mathieu-plak (commented):
A few comments / requests inline.

Also I'm kinda confused, where is the proto definition for this? 

I'm also a bit confused as how this will be used.

Peralban:
> @mathieu-plak 
> Also I'm kinda confused, where is the proto definition for this?

To be honest I don't know where put it so for the moment it is on the sdk side.
Do you want me to move it in `connectors/grpc` folders ?

connectors/grpc/storage/grpc_storage.go
landryarki: the word 'chunks' here represent chunks of the a single state or packfiles, not an actual kloset 'chunck', its a chunk of chunk if u prefer

mathieu-plak: I know that it means your protocol chunks, but when used here: 
https://github.com/PlakarKorp/plakar/pull/1099/files#diff-2b634862f67875f7883af29faca66933f75a619ead941182cb16cd670adfcca0R163

(in GetState), those Statefiles are _very_ large and on terabytes repo won't fit in memory.

connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+func toGrpcMAC(mac objects.MAC) *grpc_storage.MAC {
+	return &grpc_storage.MAC{Value: mac[:]}
+}
+
+func (s *GrpcStorage) GetStates() ([]objects.MAC, error) {
+	resp, err := s.GrpcClient.GetStates(s.ctx, &grpc_storage.GetStatesRequest{})
+	if err != nil {
+		return nil, err
```
mathieu-plak: Please wrap this error like below.

connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	var states []objects.MAC
+	for _, mac := range resp.Macs {
```
mathieu-plak, line 165: We might need to length checks the mac here to avoid a rogue plugin panic'ing plakar.

connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,349 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to delete state: %w", err)
+	}
+	return nil
+}
+
+func (s *GrpcStorage) GetPackfiles() ([]objects.MAC, error) {
+	resp, err := s.GrpcClient.GetPackfiles(s.ctx, &grpc_storage.GetPackfilesRequest{})
+	if err != nil {
+		return nil, err
```
mathieu-plak: Wrap error here.

connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,349 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to delete packfile: %w", err)
+	}
+	return nil
+}
+
+func (s *GrpcStorage) GetLocks() ([]objects.MAC, error) {
+	resp, err := s.GrpcClient.GetLocks(s.ctx, &grpc_storage.GetLocksRequest{})
+	if err != nil {
+		return nil, err
```
mathieu-plak: Wrap error here.

connectors/grpc/storage/grpc_storage.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+func (g *grpcChunkReader) Read(p []byte) (int, error) {
+	totalRead := 0
+	for totalRead < len(p) {
+		//if there is data in the internal buffer -> read from it first
+		if g.buf.Len() > 0 {
+			n, err := g.buf.Read(p[totalRead:])
+			totalRead += n
+			if err != nil && err != io.EOF {
```
mathieu-plak: err can not been EOF here.

mathieu-plak (changes requested):

PR #1100, opened 2025-06-05
stdio: make the importer work over the agent

omar-polo:


omar-polo:
kloset pr: https://github.com/PlakarKorp/kloset/pull/20

agent/agent.go
```diff
@@ -122,6 +123,21 @@ func (c *Client) SendCommand(ctx *appcontext.AppContext, name []string, cmd subc
 			return 1, fmt.Errorf("failed to decode response: %w", err)
 		}
 		switch response.Type {
+		case "stdin":
+			var buf [1024]byte
```
poolpOrg: any reason to buffer only 1k rather than 8k which is common for BUFSIZ ?

omar-polo: fixed

PR #1102, opened 2025-06-06
improve config commands

ericfaurot:
replace "create" subcommand with "add"
add "rm" subcommand
add "ls" subcommand
implement "check" and "ping"


subcommands/config/kloset.go
```diff
@@ -68,28 +72,93 @@ func cmd_kloset_config(ctx *appcontext.AppContext, args []string) error {
...
+		}
+		var list []string
+		for name, _ := range ctx.Config.Repositories {
+			list = append(list, name)
+		}
+		sort.Strings(list)
+		pfx := ""
+		for _, name := range list {
+			cfg := ctx.Config.Repositories[name]
+			fmt.Printf("%s[%s]\nlocation=%s\n", pfx, name, cfg["location"])
```
mathieu-plak: No Printf, let's use Ctx.Stdout.

subcommands/config/kloset.go
```diff
@@ -68,28 +72,102 @@ func cmd_kloset_config(ctx *appcontext.AppContext, args []string) error {
...
+		}
+		name := args[0]
+		if !ctx.Config.HasRepository(name) {
+			return fmt.Errorf("kloset %q does not exists", name)
+		}
+		_, err := storage.New(ctx.GetInner(), ctx.Config.Repositories[name])
+		if err != nil {
+			return err
+		}
+		return nil
```
omar-polo: nit: we should close the store returned

subcommands/config/kloset.go
```diff
@@ -68,28 +72,102 @@ func cmd_kloset_config(ctx *appcontext.AppContext, args []string) error {
...
+		}
+		name := args[0]
+		if !ctx.Config.HasRepository(name) {
+			return fmt.Errorf("kloset %q does not exists", name)
+		}
+		_, _, err := storage.Open(ctx.GetInner(), ctx.Config.Repositories[name])
+		if err != nil {
+			return err
+		}
+		return nil
```
omar-polo: ditto

subcommands/config/remote.go
```diff
@@ -68,16 +72,87 @@ func cmd_remote_config(ctx *appcontext.AppContext, args []string) error {
...
+		}
+		name := args[0]
+		if !ctx.Config.HasRemote(name) {
+			return fmt.Errorf("kloset %q does not exists", name)
+		}
+		_, err := storage.New(ctx.GetInner(), ctx.Config.Remotes[name])
+		if err != nil {
+			return err
+		}
+		return nil
```
omar-polo: this looks wrong to me; this is not a storage config but an importer and/or exporter one, so we should try to instantiate that.  There's no guaranteed that a remote (which we might rename connector?) is also a storage; take ftp for instance: we can backup/restore it but we can't currently have a repository over bare ftp.  (or think of the work the interns are doing, we're not going to store a kloset in, say, notion or icloud photo)

subcommands/config/remote.go
```diff
@@ -68,16 +72,87 @@ func cmd_remote_config(ctx *appcontext.AppContext, args []string) error {
...
+		}
+		name := args[0]
+		if !ctx.Config.HasRemote(name) {
+			return fmt.Errorf("remote %q does not exists", name)
+		}
+		_, _, err := storage.Open(ctx.GetInner(), ctx.Config.Remotes[name])
+		if err != nil {
+			return err
+		}
+		return nil
```
omar-polo: ditto

omar-polo (commented):
two doubts:
1. what's the advantage to have both check and ping?  one of the two is not enough?  (also because the only difference is that in ping we deserialize the config, in check not)
2. the ping/check implementation for remotes is wrong, we can't open a storage with the params of a remote, if it works it's by chance.

poolpOrg (approved):

mathieu-plak (approved):

PR #1103, opened 2025-06-06
FEAT: Create load functions for plugins

Peralban:
I hate code duplication, its why this function looks like that.
All comment are temporary. It will be deleted one by one when the grpc structs will be merged

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -124,6 +124,8 @@ require (
 	golang.org/x/net v0.38.0 // indirect
 	golang.org/x/sys v0.31.0 // indirect
 	golang.org/x/text v0.23.0 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20250603155806-513f23925822 // indirect
+	google.golang.org/grpc v1.73.0 // indirect
 	google.golang.org/protobuf v1.36.6 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	modernc.org/libc v1.62.0 // indirect
--- a/main.go
+++ b/main.go
@@ -28,6 +28,7 @@ import (
 	"github.com/PlakarKorp/plakar/subcommands"
 	"github.com/PlakarKorp/plakar/task"
 	"github.com/PlakarKorp/plakar/utils"
+	"github.com/PlakarKorp/plakar/plugins"
 	"github.com/denisbrodbeck/machineid"
 	"github.com/google/uuid"
 
@@ -158,6 +159,8 @@ func EntryPoint() int {
 
 	ctx := appcontext.NewAppContext()
 	defer ctx.Close()
+
+	plugins.LoadBackends(ctx.Context, filepath.Join(configDir, "plugins"))
 
 	cfg, err := config.LoadOrCreate(opt_configfile)
 	if err != nil {
--- a/plugins/plugin_loader.go
+++ b/plugins/plugin_loader.go
@@ -1,0 +1,193 @@
+package plugins
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"os"
+	"path/filepath"
+	"regexp"
+	"syscall"
+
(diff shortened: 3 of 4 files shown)


go.mod
```diff
@@ -124,6 +124,8 @@ require (
 	golang.org/x/net v0.38.0 // indirect
 	golang.org/x/sys v0.31.0 // indirect
 	golang.org/x/text v0.23.0 // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20250603155806-513f23925822 // indirect
+	google.golang.org/grpc v1.73.0 // indirect
```
omar-polo: nit, i guess you have to re-run `go mod tidy` since these are not "indirect" dependencies

main.go
```diff
@@ -159,6 +160,8 @@ func EntryPoint() int {
 	ctx := appcontext.NewAppContext()
 	defer ctx.Close()
 
+	plugins.LoadBackends(ctx.Context, filepath.Join(configDir, "plugins"))
```
omar-polo: nit: we'd have to do something if this fails

plugins/plugin_loader.go
```diff
@@ -0,0 +1,193 @@
...
+		return -1, -1, fmt.Errorf("failed to create socketpair: %w", err)
+	}
+
+	procAttr := syscall.ProcAttr{}
+	procAttr.Files = []uintptr{
+		os.Stdin.Fd(),
+		os.Stdout.Fd(),
+		os.Stderr.Fd(),
+		uintptr(sp[0]),
+	}
```
omar-polo: why fd 4 for the pipe?  i'm not sure that's portable.

plugins/plugin_loader.go
```diff
@@ -0,0 +1,193 @@
...
+
+	var pid int
+
+	argv := []string{pluginPath, fmt.Sprintf("%v", config)}
+	pid, err = syscall.ForkExec(pluginPath, argv, &procAttr)
+	if err != nil {
+		return -1, -1, fmt.Errorf("failed to ForkExec: %w", err)
+	}
```
omar-polo: can't we use [`os/exec.Cmd`](https://pkg.go.dev/os/exec#Cmd) instead?  it would work out of the box on windows as well :)

omar-polo (changes requested):
this is just a first round of review, commenting on some superficial stuff.

i still need to grok the main part of it, will need a little bit more time.

plugins/plugin_loader.go
Peralban: I ve done it with @poolpOrg I think he can have a better answer than me.

plugins/plugin_loader.go
Peralban: Same I'm not familiar with that function maybe ask to @poolpOrg :)

plugins/plugin_loader.go
```diff
@@ -0,0 +1,145 @@
...
+	cmd.Stdin = childFile
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	if err := cmd.Start(); err != nil {
+		return -1, fmt.Errorf("failed to start plugin: %w", err)
+	}
+
+	childFile.Close()
+	return int(sp[1]), nil
```
poolpOrg: why cast ? it's already an int

plugins/plugin_loader.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil, err
+	}
+
+	connFile := os.NewFile(uintptr(fd), "grpc-conn")
+	conn, err := net.FileConn(connFile)
+	if err != nil {
+		return nil, fmt.Errorf("net.FileConn failed: %w", err)
+	}
+
+	clientConn, err := grpc.NewClient("localhost:0",
```
poolpOrg: should be 127.0.0.1:0, not localhost:0

PR #1115, opened 2025-06-12
importer/fs: add knob to not traverse filesystems

omar-polo:
This adds a knob for the fs importer to not traverse filesystems during a backup.  The idea is that we keep track of the device number for the root of the backup and we skip all the directories that belong to a different device.

The new knob is called dont_traverse_fs and at the moment has to be specified by creating a "remote" config stanza with a trueish value ("1", "t", "T", "true", "TRUE" or "True").

### DIFF
--- a/connectors/fs/importer/fs.go
+++ b/connectors/fs/importer/fs.go
@@ -24,8 +24,10 @@ import (
 	"os/user"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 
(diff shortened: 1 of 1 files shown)


omar-polo:
(this pr is built on top of `op/madness` because i needed the workspace fix as well to work.)

omar-polo:
with https://github.com/PlakarKorp/plakar/pull/1116 this would just be `plakar backup -o dont_traverse_fs=true /`

mathieu-plak (approved):

PR #1121, opened 2025-06-16
move proto

landryarki:
proto from sdk repo are now here


Makefile
```diff
@@ -25,4 +25,28 @@ check: test
...
+HOME := $(shell echo $$HOME)/
+
+gen:
+	for proto in $(PROTOS); do \
+		cp connectors/grpc/$$proto/$$proto.proto .; \
+		mkdir -p ./pkg/$$proto/; \
+		docker run --rm -ti \
+			-v `pwd`:/app \
+			-w /app \
+			rvolosatovs/protoc \
```
omar-polo: hugh, really? :(

Makefile
```diff
@@ -25,4 +25,28 @@ check: test
 test:
 	${GO} test ./...
 
-.PHONY: all plakar install check test
+PROTO_SRC_DIR := $(shell pwd)/proto
```
omar-polo: not used, can be removed

Makefile
```diff
@@ -25,4 +25,28 @@ check: test
...
+PROTOS := importer exporter storage
```
omar-polo: no need to use `:=` in here, just `PROTOS = importer exporter storage` is fine

Makefile
```diff
@@ -25,4 +25,28 @@ check: test
...
+HOME := $(shell echo $$HOME)/
```
omar-polo: unused

Makefile
omar-polo: can we avoid running docker in here?  protoc is famous enough that you can get it from your favourite package manager instead of running some random docker image.

omar-polo (approved):
much better, thank you!

can't comment much on the protobuf files themselves, but let's get them into the tree and improve on top later.

mathieu-plak (approved):

Issue #1058, opened 2025-05-17
Windows: 1st try to compile


keltia:
As I use restic for backup-ing my windows machine (games & dev), I tried plakar for the first time there too (hello Gilles :) )

```text
go: downloading github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec
# github.com/PlakarKorp/plakar/cmd/plakar/subcommands/agent
..\..\..\go\pkg\mod\github.com\!plakar!korp\plakar@v1.0.1\cmd\plakar\subcommands\agent\agent.go:89:22: undefined: syscall.ForkExec
..\..\..\go\pkg\mod\github.com\!plakar!korp\plakar@v1.0.1\cmd\plakar\subcommands\agent\agent.go:126:20: undefined: syslog.New
..\..\..\go\pkg\mod\github.com\!plakar!korp\plakar@v1.0.1\cmd\plakar\subcommands\agent\agent.go:126:31: undefined: syslog.LOG_INFO
..\..\..\go\pkg\mod\github.com\!plakar!korp\plakar@v1.0.1\cmd\plakar\subcommands\agent\agent.go:126:47: undefined: syslog.LOG_USER
..\..\..\go\pkg\mod\github.com\!plakar!korp\plakar@v1.0.1\cmd\plakar\subcommands\agent\agent.go:171:10: undefined: syscall.Kill
```

poolpOrg:
Hi @keltia !

Thanks for the report, fyi the Windows port is still in progress, we have managed to build, run and exchange snapshots between Unix and Windows (and the other way around) but there's still a bit of work to do before it's usable.

It's likely to be ready in the next few weeks, I'll notify you on this issue when you can give it a try.

keltia:
Lemme know if I can help BTW.

omar-polo:
a small update; i managed to make my first backup on windows today!

I'm starting to clean up my hacky branch into smaller self-contained pull requests.  There is still more to adjust for a on-par user experience, but I'm making progresses :)

poolpOrg:
@keltia it's not fully finished but it builds and runs thanks to @omar-polo 's Windows Expertise.

Care to try ?

PhilJbt:
I have the same problem as @keltia with `v1.0.2` on Win10.

> Care to try ?

@poolpOrg I'd be interested in testing it, which branch/release should I go for?

omar-polo:
[...]

PR #1122, opened 2025-06-16
fix some path vs filepath usage

omar-polo:
path is used to deal with UNIX-like paths, for e.g. URLs (stretching it a bit but still), while filepath is for the *local* operating system path format.

So, stick with path for remotes like (s)ftp and stdin, while the local fs has to use filepath.


omar-polo:

mathieu-plak (approved):
LGTM

PR #1123, opened 2025-06-16
for auto-generated paths, prefix "fs:"

omar-polo:
This is meant to help with windows, where a path like "C:\Users\Omar\" might be mistaken for a remote with "C" protocol.

What to exactly do in the command line when a user types the full path with the device included is still to solve.


omar-polo:
go fmt included because why not

omar-polo:

poolpOrg (approved):
shouldn't it be fs:// out of coherency with the fact we always generate the long version ?

mathieu-plak (approved):

PR #1126, opened 2025-06-16
Fix NormalizePath for windows and unices

omar-polo:
- this fixes NormalizePath on windows.  I'm arbitrary choosing `C:\\` for the empty string case, even if it's currently impossible since NormalizePath is only called on the result of `os.Getwd`.
- then it moves NormalizePath in a different file that's built only on windows and mac, since on unix the paths are case-sensitive (well, we don't even have the concept of 'case' luckily)


omar-polo:

PR #1132, opened 2025-06-19
avoid a cross-device rename

omar-polo:
Not specifying the ConfigDir means we're creating files in the /tmp/ directory which might reside on a different file system.


ericfaurot (approved):
yes!

PR #1134, opened 2025-06-19
some quick fixes for windows

omar-polo:
These are just faking some functions on windows, they'll need to be actually implemented in a follow up, but for now it make easier to add windows to the CI and stop breaking it all the time.


omar-polo:

mathieu-plak (approved):

PR #1156, opened 2025-06-23
more windows shimming

omar-polo:


omar-polo:

(and actually finally fixes the build OOTB on windows, even if there are still some windows-specific things to do)

PR #1166, opened 2025-06-24
Introduce plakar cookies.

mathieu-plak:
* This is moving the cookie implementation and storage in plakar following their removal from kloset.

* This diff is a bit more churn than I would have liked because of the api madness and the task stuff that was not taking a context (fixed now)

* While here remove the cookiesDir from the context, the manager saves it ougth to be enough.

* Simplifies the tests, there is no need for the tests to set up a cookies manager..

* This depends on https://github.com/PlakarKorp/kloset/pull/37

### DIFF
--- a/cookies/cookies.go
+++ b/cookies/cookies.go
@@ -32,6 +32,10 @@ func (m *Manager) Close() error {
 	return nil
 }
 
+func (m *Manager) GetDir() string {
+	return m.cookiesDir
+}
+
 func (c *Manager) GetAuthToken() (string, error) {
 	data, err := os.ReadFile(filepath.Join(c.cookiesDir, ".auth-token"))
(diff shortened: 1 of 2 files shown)


mathieu-plak:
Tests are failing I have no idea why. They do run locally

omar-polo (approved):
lovely

i think the tests are failing because we have to update kloset; let's merge that other PR, update the dep and retry.

mathieu-plak:
> lovely
> 
> i think the tests are failing because we have to update kloset; let's merge that other PR, update the dep and retry.

Yep, I realised as much this morning. Was kinda confused because it fails at runtime.

poolpOrg (approved):

omar-polo (approved):

PR #1167, opened 2025-06-24
kcontext thinning

mathieu-plak:
depends on: https://github.com/PlakarKorp/kloset/pull/38

### DIFF
--- a/main.go
+++ b/main.go
@@ -166,7 +166,6 @@ func EntryPoint() int {
 
 	ctx.Client = "plakar/" + utils.GetVersion()
 	ctx.CWD = cwd
-	ctx.KeyringDir = filepath.Join(opt_userDefault.HomeDir, ".plakar-keyring")
 
 	_, envAgentLess := os.LookupEnv("PLAKAR_AGENTLESS")
 	if envAgentLess {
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -256,15 +255,13 @@ func EntryPoint() int {
...
-	ctx.NumCPU = opt_cpuCount
 	ctx.Username = opt_username
 	ctx.Hostname = opt_hostname
 	ctx.CommandLine = strings.Join(os.Args, " ")
 	ctx.MachineID = opt_machineIdDefault
 	ctx.KeyFromFile = secretFromKeyfile
-	ctx.HomeDir = opt_userDefault.HomeDir
 	ctx.ProcessID = os.Getpid()
-	ctx.MaxConcurrency = ctx.NumCPU*8 + 1
+	ctx.MaxConcurrency = opt_cpuCount*8 + 1
```
omar-polo, line 264: it was a problem even before i guess but what if i do `-cpu -8`?  should we guard that it must be zero at least?

mathieu-plak: I guess yeah. Will send a followup PR.

poolpOrg (approved):

PR #1169, opened 2025-06-24
backup: Fix @source notation.

mathieu-plak:
* Source notation is broken in backup because when we merge options from the config we give precedence (rightfully) to the command line provided options but we also _always_ fixup the location in the options before doing this, which means we never let the conf get applied.

* Note that there is no way to specify a location if we are resolving a "source" since... we use location to specify the source so this is safe to do (aka we will never _not_ apply a CLI param for the location)

* found by @brmzkw  again! Thx!

### DIFF
--- a/subcommands/backup/backup.go
+++ b/subcommands/backup/backup.go
@@ -170,10 +170,6 @@ func (cmd *Backup) DoBackup(ctx *appcontext.AppContext, repo *repository.Reposit
 		scanDir = cmd.Path
 	}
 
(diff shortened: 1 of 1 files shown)


omar-polo (approved):

brmzkw:
This PR is fixing the following:

```bash
./plakar at ./toto backup @minio
./plakar: failed to create an importer for @minio: lstat /Users/niluje/dev/plakar/plakar/@minio: no such file or directory
```

PR #1197, opened 2025-06-30
pkg create: add extra_files to the manifest; improve usage

omar-polo:
- now the manifest has to list all the files
- all files listed by the manifest must be at or below the manifest directory
- ensure that the executables are, in fact, executable


omar-polo:
bummer that i had to copy the stdio event processor from backup, in a follow-up we may dedup those.

PR #1212, opened 2025-07-02
adapt to the kloset change in api

poolpOrg:


mathieu-plak (approved):
yes but why?

omar-polo (approved):
needs a go get, otherwise ok for me

PR #1220, opened 2025-07-03
pkg subcommand improvements

ericfaurot:
- remove the installed ptar if loading fails during install
- add a "pkg ls" subcommands to display installed plugins
- add a "pkg uninstall" subcommand

### DIFF
--- a/subcommands/pkg/uninstall.go
+++ b/subcommands/pkg/uninstall.go
@@ -1,0 +1,95 @@
+/*
+ * Copyright (c) 2025 Eric Faurot <eric.faurot@plakar.io>
+ *
+ * Permission to use, copy, modify, and distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+ * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ * ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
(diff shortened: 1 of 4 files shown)


subcommands/pkg/uninstall.go
```diff
@@ -0,0 +1,95 @@
...
+	pluginsDir := filepath.Join(dataDir, "plugins")
+	names, err := plugins.ListDir(ctx, pluginsDir)
+	if err != nil {
+		return 1, err
+	}
+
+	for _, name := range cmd.Args {
+		if !slices.Contains(names, name) {
+			return 1, fmt.Errorf("plugin %q is not installed", name)
+		}
```
omar-polo, line 81: nit: instead of doing a List, I'd just attempt at `os.Remove` the plugin and then check whether the error returned is ENOENT to warn "it's not installed".

subcommands/pkg/uninstall.go
```diff
@@ -0,0 +1,95 @@
...
+		err := os.Remove(filepath.Join(pluginsDir, name))
+		if err != nil {
+			return 1, fmt.Errorf("failed to remove %q: %w", name, err)
+		}
+		extlen := len(filepath.Ext(name))
+		pluginCache := filepath.Join(cachedir, name[:len(name)-extlen])
+		err = os.RemoveAll(pluginCache)
+		if err != nil {
+			return 1, fmt.Errorf("failed to remove cache for %q: %w", name, err)
+		}
```
omar-polo, line 91: same here, i'd make this silent is the directory doesn't exist.

omar-polo (approved):
looks good to me, thanks for this!  just left a small nitpick below, but i can fix it on another PR if you want (i was also about to move some code from the pkg command line back in the plugins/ lib.)

subcommands/pkg/uninstall.go
poolpOrg: yes

subcommands/pkg/uninstall.go
poolpOrg: agreed too

poolpOrg (approved):

subcommands/pkg/uninstall.go
mathieu-plak: Will be fixed by @omar-polo  in a follow up PR.

subcommands/pkg/uninstall.go
mathieu-plak: Will be fixed by @omar-polo  in a follow up PR.

PR #1225, opened 2025-07-03
Improve plakar help backup

brmzkw:
- update last review date
- document silent option
- similarly to c2cdb528, "importer" is now "source of the integration" and "repository" is "Kloset store"
- plakar-config(1) no longer exists

PR #1226, opened 2025-07-05
Allow several tags in plakar backup

brmzkw:
If you merge this PR, it will merge the code to `jcastets/plakar-backup-doc`.

This PR must be merged before https://github.com/PlakarKorp/plakar/pull/1225

### DIFF
--- a/subcommands/backup/backup.go
+++ b/subcommands/backup/backup.go
@@ -49,9 +49,31 @@ func (e *excludeFlags) Set(value string) error {
 	return nil
 }
 
+type tagFlags string
+
+// Called by the flag package to print the default / help.
+func (e *tagFlags) String() string {
+	return string(*e)
+}
+
+// Called once per flag occurrence to set the value.
+func (e *tagFlags) Set(value string) error {
+	if *e != "" {
+		return fmt.Errorf("tags should be specified only once, as a comma-separated list")
+	}
+	*e = tagFlags(value)
+	return nil
+}
+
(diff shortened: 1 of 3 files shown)


subcommands/backup/backup.go
```diff
@@ -38,20 +38,24 @@ func init() {
...
 }
 
-type excludeFlags []string
+type flagList []string
 
-func (e *excludeFlags) String() string {
+// Called by the flag package to print the default / help.
+func (e *flagList) String() string {
 	return strings.Join(*e, ",")
 }
```
omar-polo: i thought we were going to go with the `-tag one,two,three` instead of `-tag one -tag two -tag three`, but anyway it's OK for me.  My only "gripe" is that if we go with multiple -tags then this is wrong, but on the other hand i don't see how to fix it.

just wanted to point it out, not a big deal.

omar-polo (approved):

subcommands/backup/backup.go
poolpOrg: I really think we shoudl go -tag one,two,three with exclusive -tag (option can't be passed twice otherwise I can _always_ come with a way to shoot myself in the foot)

poolpOrg:
UNAPPROVED

brmzkw:
Screenshot 2025-07-07 at 23 03 45

Fixed

poolpOrg (commented):
OK

PR #1229, opened 2025-07-05
CLI-side, if no path was provided, pass ctx.CWD as cmd.Path

poolpOrg:


omar-polo (commented):
from some initial testing, this seems to just work around the issue.  The problem is that ctx.CWD in Execute is not the cwd of the client but the one of the agent

omar-polo:
because as far as i can see there is at least another place where we use `ctx.CWD` in a Execute function and that is in restore, where we have the same bug, and in the pkg tooling if we decide to run them on the agent (currently that's not the case.)

in https://github.com/PlakarKorp/plakar/pull/1234 I'm attempting to fix this by (ab)using the SubcommandBase struct to pass the CWD thru the socket and deserialize it, but not sure that's the best option, it was just the quickest i could think.

poolpOrg:
I explained my rationale on discord but should have done it here:

I think we should guarantee that CWD is correctly reflected on the agent side, but I'm unsure we should rely on this when we can actually enforce a parameter on the CLI-side.

To me we should have both the CLI-side pass explicit paramters to rpc, and the agent side have a correct context.

omar-polo:
now i understand; apologize but i must have missed the message on discord.  Then i'm all in for this and I agree, I was only reasoning from the context point of view.

omar-polo (approved):

PR #1232, opened 2025-07-05
Add Docker configuration and startup script for Plakar UI

guigui42:
The goal is to make it easy to self host plakar ui using Docker.

### Docker support for Plakar UI:

* **`.dockerignore`**: Added a `.dockerignore` file to exclude unnecessary files from the build context, such as Git files, documentation, test files, CI/CD files, and temporary files.
* **`Dockerfile`**: Created a multi-stage Dockerfile to build and run the Plakar UI application. The build stage compiles the application using Go, while the runtime stage sets up a minimal Alpine-based environment with a non-root user for security, health checks, and default startup commands.
* **`docker/docker-compose.yml`**: Added a Docker Compose configuration to simplify running the Plakar UI in a container. It includes service definitions, port mappings, volume mounting for the repository, environment variable configuration, and automatic restart policies.
* **`docker/start-plakar-ui.sh`**: Introduced a startup script to launch the Plakar agent and UI server within the container. The script ensures proper initialization and error handling.

Next step would be to create a Workflow to build the image and push to GitHub directly and make the docker image available to everyone whenever there is a new Release. (if needed I can also take care of it )


docker/start-plakar-ui.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Check if agent is still running
+if ! kill -0 $AGENT_PID 2>/dev/null; then
+    echo "Agent failed to start"
+    exit 1
+fi
+
+echo "Starting Plakar UI..."
+# Start the UI server with passed arguments
+exec plakar ui "$@"
```
omar-polo: starting the agent in this case is pointless, since we're only going to have one plakar instance running on the container, so we can avoid this altogether and just run `plakar -no-agent ui` in the dockerfile.

Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+COPY . .
+
+# Build the application
+RUN make plakar
+
+# Final stage - minimal runtime image
+FROM alpine:latest
+
+# Install required runtime dependencies
+RUN apk add --no-cache ca-certificates tzdata bash
```
omar-polo: do we really need bash?  even if we keep the wrapper script, that looks like a POSIX sh script to me, so we can avoid the bash dependency.

omar-polo (commented):
hello, thanks for the contribution.

I haven't tested this yet, will do it soon, i've left a first round of comments on it though.

docker/start-plakar-ui.sh
guigui42: Makes sense, didnt' see that option. custom script remove and command line simplified.

Dockerfile
guigui42: dependency removed

omar-polo: thank you!

docker/start-plakar-ui.sh
omar-polo: great, thanks :)

Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+# Set environment variables
+ENV PLAKAR_REPOSITORY=/data
+ENV PLAKAR_PASSPHRASE=""
+
+# Health check to ensure the service is running
+HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
+  CMD wget --no-verbose --tries=1 --spider http://localhost:8080/ || exit 1
+
+# Default command to start the UI with agent
+# Users can override with custom arguments
```
omar-polo: last nit while you're at it, the comment is now outdated since it no longer runs the agent.

omar-polo (approved):
one last nit but overall this looks good to me.  We might want to refactor a little bit in tree, since it could be handy to have an plakar image to do other things beside running the ui, like automatize backups in certain scenarios, but for now this is OK for me.

Thank you! :)

guigui42:
Question : for Docker images in the future, would you want to host them on GitHub directly ? I may be biased toward GitHub ;)

omar-polo:
really have to be discussed (+cc @poolpOrg) but for starting using the gh artifact registry seems fine to me, and also quite easy ;-)

Dockerfile
guigui42: Done. 
What kind of refactor are you thinking about ?

omar-polo:
@guigui42 
> What kind of refactor are you thinking about ?

well, to be discussed honestly, but I'd  remove the UI specific things from the dockerfile, so the resulting image is easier to use for backup tasks as well, and move them to the compose file.

poolpOrg:
we invoke the @brmzkw for comment

poolpOrg:
After team discussions I'm not convinced about the UI docker thing, I see value in having the beginning of a Docker project (a skel of some sort) but not the ui stuff, I need some more thinking before I approve

brmzkw:
Hello,

First of all, thanks **a lot** for your contribution.

However, I don’t think it should be merged because I’m not sure I see the value of adding a way to run Plakar in a Docker container the way it’s done here. Let me explain why, and please, if you disagree, I’m completely open to changing my mind.

Docker containers are useful to run software in an isolated, standalone environment. If you wanted to run, for example, an FTP server, it makes complete sense to host it in a Docker container to avoid installing all the dependencies on the host.

For Plakar, I believe things are a bit different.

Kloset stores support various storage connectors. They can be hosted on the filesystem, in an S3 bucket, on an SFTP server, and so on.

In the Dockerfile and docker-compose.yml files of this PR, we assume the Kloset store is mounted in /data, which implies:
* The Kloset store must exist on the host — which means Plakar is installed on the host.
* There’s no way to use this image to serve a Kloset store hosted on S3, SFTP, or any other storage connector, since the Kloset store is mounted as a Docker volume.

Since Plakar must be installed on the host, why would you run it in a Docker container?

I see two interesting alternatives for running Plakar in a Docker container:
1. Create a Docker image that connects to a Plakar server. A Plakar server exposes the Kloset store over HTTP — think of it as a proxy. This way, you can run plakar server on any host, serving a Kloset repository using any storage connector, and run a Docker container elsewhere that uses this server.
2. Alternatively, create a custom Docker volume driver that connects to a Plakar server and exposes the Docker volume.

For option 2, I have a proof of concept that's **far** from being ready for release, but it looks like this:

* Create a Docker volume:
```bash
docker plugin install brmzkw/kloset --grant-all-permissions
docker volume create myvolume -d brmzkw/kloset
```

* From another shell, start the Plakar server:
```bash
plakar at ./my_kloset server
```

* Then start a container:
```bash
docker run --rm -ti -v myvolume:/data ubuntu bash
```

This way, your Kloset stored is exposed at `/data`.


===

Here, I’ve tried to explain my thoughts on Docker and Plakar. Again, I'm completely open to changing my mind if there are use cases I haven’t thought of.

guigui42:
@brmzkw That is a very clear explanation, thanks.

My original goal was to easily setup my Traefik reverse proxy, that runs on Docker alongside other web UIs for other tools.

In my case, Plakar is not necessarily installed on my docker server,  the Plakar / Kloset is on a shared mounted folder (used by other servers on their own mounted folder).

Having Plakar UI on docker is an easy way for me to manage the web interface, start it, stop it from Portainer, as well as including it in my Traefik configuration, ssl certificates ...

I could of course install & run Plakar on the docker server as well, I don't have any isolation requirements, it is just a configuration / management convenience.

For the /data folder, there are indeed better ways to do it in order to handle more than just file-system folders, it just wasn't in my initial scope since I only use "fs" for now.

To sum up, my usage is very much for a Home Setup (trying to replace Synology tools) and may vary from other users I understand.
Feel free to close the PR if it doesn't fit the Plakar vision.

I am still discovering Plakar and already very happy with it even if I have to build my own docker image ;) 
So thanks for all your work on it !

maxime1992:
> Since Plakar must be installed on the host, why would you run it in a Docker container?

Why would you need to have Plakar installed on the host?

It's very convenient to be able to run everything into docker and if you look at alternatives like Duplicati, Kopia and others they all offer that option. I was about to give Plakar a go but not having a docker image is currently a no go for me.

brmzkw:
Don’t get me wrong, I’m not opposed to including a Dockerfile.

> Why would you need to have Plakar installed on the host?

With this specific Dockerfile, the Kloset store is expected to be mounted as a Docker volume at `/data`, which means that volume must already exist on the host. To create it, Plakar needs to have been installed there.

I'm all for providing a Dockerfile, but I don't think this particular one belongs in our repository. I'd prefer something that supports more than just FS Kloset stores, so I'd rather wait a bit before including it.

I do understand the value of a Dockerfile. Personally, I use Scaleway's serverless services to run Plakar every day for backups. My main issue is that the ephemeral container starts with an empty cache. While an empty cache isn't a problem per se, it does make Plakar less efficient. In my own image, I manually persist the cache after each backup and restore it before the next run. I'd love for us to provide a generic solution for cache persistence in the official Docker image.

benoit2600:
Hello,

I'd like to present my use case for Plakar, and why a Dockerfile with UI access would be useful to me.

I want to back up data from my PC to my NAS (a proprietary QNAP NAS).

Therefore, I plan to install Plakar on my PC and send the data via SSH to the NAS.

Now, if I'm at work or a friend's house and I need to access a file from my PC, but my PC is off, I should be able to use one of the snapshots present on the NAS to do so.

The problem is, I don't have Plakar on the NAS. I also can't install it at work, and it's always tricky to install it on a friend's PC.

If I had the web interface accessible from the NAS, I could browse my data. Since my NAS is proprietary, I can't directly install Go and Plakar; I need to use a container or a Dockerfile.

brmzkw:
Thanks for sharing your use-case.

We will definitely work on a Docker image, but I think we'll wait for the remote cache first. I'm scared that if we provide a Docker image now, people will use it and we'll have a lot of support requests because the cache wouldn't work as expected.

Until then, you can use the following Dockerfile:

```dockerfile
FROM golang

RUN go install github.com/PlakarKorp/plakar@latest

VOLUME /var/kloset
EXPOSE 8080

CMD ["/go/bin/plakar", "-no-agent", "at", "/var/kloset", "ui", "-addr", "0.0.0.0:8080", "-no-spawn", "-no-auth"]
```

And use it as follow:

```bash
docker build -t my-plakar-docker-image .
docker run -v ~/.plakar:/var/kloset --rm -ti -p 8080:8080 my-plakar-docker-image
```

The UI is accessible at localhost:8080, and serves the Kloset store from the host at `~/.plakar`.

thomas-illiet:
@guigui42 Please update the Go version to match the one specified in go.mod :)

Otherwise, if we assume that security needs to be pushed to 100% (for maximum security, of course 😄), the Dockerfile would need to be modified, as the following volumes are missing:

- /home/appuser/.config
- /home/appuser/.cache
- /home/appuser/.local

Here is the command I’m using to test the product:

```bash
docker run -it --rm \
    --name plakar-in-blackhole \
    --cap-drop=ALL \
    --security-opt no-new-privileges:true \
    --user 1000:1000 \
    --read-only \
   -p 8085:8085 \ 
   --volume /appdata/plakar/config:/home/appuser/.config \
   --volume /appdata/plakar/cache:/home/appuser/.cache  \
   --volume /appdata/plakar/local:/home/appuser/.local \
   --volume /appdata/plakar/data:/data \
   ghcr.io/myrepository/plakar:latest \
   bash
```

brmzkw:
Closing, since the Dockerfile has been merged a while ago.

PR #1234, opened 2025-07-06
fix ctx.CWD thru the agent

omar-polo:


omar-polo:
this lacks a regression suite added, i'm not sure how to simulate this case currently but will try during the week.

PR #1240, opened 2025-07-07
 teach pkg install how to fetch from http

omar-polo:
WIP: i still have my "temp" endpoint for testing in there because the infra is still not available


poolpOrg (approved):
OK

mathieu-plak (approved):

PR #1248, opened 2025-07-08
Fix .vscode/launch.json

brmzkw:
The path to the debugger has not been updated when we moved plakar to the root folder.


poolpOrg (approved):

mathieu-plak (approved):
wasn't even aware this was a thing.

PR #1257, opened 2025-07-08
sftp: Fix leak of files.

mathieu-plak:
* Since kloset switched from io.ReadAll() to io.ReadFull() we have been leaking file. Our ClosingFileReader is basing the end of read detection on EOF, but io.ReadFull doesn't emit an EOF (rightfully) so we would just never close teh underlying object.

* Since we can't mix the semantics (close on EOF / close on N bytes read) introduce a new type aptly named ClosingLimitedFileReader that will close the reader after N bytes have been read (or return an error in case of an early EOF).

* Found thanks to OpenBSD pretty aggressive limits on normal users.

### DIFF
--- a/connectors/sftp/storage/utils.go
+++ b/connectors/sftp/storage/utils.go
@@ -31,6 +31,14 @@ type ClosingFileReader struct {
 	fileClosed bool
 }
 
+type ClosingLimitedFileReader struct {
+	reader     io.Reader
+	file       *sftp.File
+	fileClosed bool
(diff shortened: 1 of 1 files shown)


ericfaurot (approved):
ok

poolpOrg:
@ericfaurot 's diff was merged in kloset and plakar, is this PR still necessary ?

mathieu-plak:
Better fix in #1265

PR #1277, opened 2025-07-10
sftp: Use ReadFromWithConcurrency for writes.

mathieu-plak:
* Using ReadFromWithConcurrency, one saves us a temporary buffer, but also let's the pkg/sftp package pipeline the send giving way more throughtput.


poolpOrg (approved):
ok !

decarheil (approved):

PR #1282, opened 2025-07-10
ptar creation with no param should do similar to backup with no param => CWD

poolpOrg:


subcommands/ptar/ptar.go
```diff
@@ -87,7 +87,8 @@ func (cmd *Ptar) Parse(ctx *appcontext.AppContext, args []string) error {
 	}
 
 	if len(cmd.SyncTargets) == 0 && flags.NArg() == 0 {
-		return fmt.Errorf("%s: at least one -k option or path must be specified", flag.CommandLine.Name())
+		cmd.BackupTargets = make([]string, 1)
+		cmd.BackupTargets[0] = ctx.CWD
```
omar-polo: style nit, i'd write it as 

```go
cmd.BackupTargets = []string{ctx.CWD}
```

which saves a linee, but OK anyway

mathieu-plak (approved):

subcommands/ptar/ptar.go
poolpOrg: solved !

PR #1283, opened 2025-07-11
backup: add GNU-style long options for flags

lcheylus:
For `backup` subcommand, use GNU-style long options and one-letter shorthands for flags using [github.com/spf13/pflag](https://github.com/spf13/pflag) package.

- Sync manpage and Markdown doc for `backup` subcommand
- Fix test with `--exclude` in `subcommands/backup/backup_test.go`

**Tests OK** for `backup` subcommand
```bash
$ go test -v github.com/PlakarKorp/plakar/subcommands/backup
=== RUN   TestExecuteCmdCreateDefault
--- PASS: TestExecuteCmdCreateDefault (0.13s)
=== RUN   TestExecuteCmdCreateDefaultWithExcludes
--- PASS: TestExecuteCmdCreateDefaultWithExcludes (0.14s)
PASS
ok      github.com/PlakarKorp/plakar/subcommands/backup 0.281s
```

`backup` subcommand help with `-h / --help` flag (disable sorting of flags):
```bash
$ ./plakar at <PATH> backup -h
Usage: backup [OPTIONS] path
       backup [OPTIONS] @LOCATION

OPTIONS:
  -x, --exclude strings       glob pattern to exclude files, can be specified multiple times to add several exclusion patterns
      --exclude-file string   path to a file containing newline-separated regex patterns, treated as --exclude
  -t, --tag strings           comma-separated list of tags to apply to the snapshot
  -n, --scan                  do not actually perform a backup, just list the files
  -q, --quiet                 suppress output
  -s, --silent                suppress ALL output
  -c, --check                 check the snapshot after creating it
      --concurrency uint      maximum number of parallel tasks (default 57)
  -o, --option option=value   specify extra importer options
      --stdio                 output one line per file to stdout instead of the default interactive output
  -h, --help                  show this help message

Run 'plakar help backup' for details.
```

### DIFF
--- a/subcommands/backup/backup.go
+++ b/subcommands/backup/backup.go
@@ -18,10 +18,11 @@ package backup
 
 import (
 	"bufio"
-	"flag"
 	"fmt"
 	"os"
 	"strings"
+
+	flag "github.com/spf13/pflag"
 
 	"github.com/PlakarKorp/kloset/objects"
 	"github.com/PlakarKorp/kloset/repository"
@@ -38,41 +39,9 @@ func init() {
 	subcommands.Register(func() subcommands.Subcommand { return &Backup{} }, subcommands.AgentSupport, "backup")
 }
 
-type excludeFlags []string
-
-func (e *excludeFlags) String() string {
-	return strings.Join(*e, ",")
-}
-
-func (e *excludeFlags) Set(value string) error {
-	*e = append(*e, value)
-	return nil
-}
-
-type tagFlags string
-
-// Called by the flag package to print the default / help.
-func (e *tagFlags) String() string {
-	return string(*e)
-}
-
-// Called once per flag occurrence to set the value.
-func (e *tagFlags) Set(value string) error {
-	if *e != "" {
-		return fmt.Errorf("tags should be specified only once, as a comma-separated list")
-	}
-	*e = tagFlags(value)
-	return nil
-}
-
-func (e *tagFlags) asList() []string {
-	return strings.Split(string(*e), ",")
-}
-
 func (cmd *Backup) Parse(ctx *appcontext.AppContext, args []string) error {
 	var opt_exclude_file string
-	var opt_exclude excludeFlags
-	var opt_tags tagFlags
+	excludeFlags := []string{}
 
 	excludes := []string{}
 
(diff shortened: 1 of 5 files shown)


poolpOrg:
If we switch to spf13, I would really prefer:

https://github.com/charmbracelet/fang

lcheylus:
@poolpOrg I agree but fang is experimental, have no detailed documentation and requires [Cobra](https://github.com/spf13/cobra) (not used by Plakar for now).

It's a more difficult task to replace all code for subcommands in Plakar using Cobra.

poolpOrg:
need a bit more thinking, but one thing for sure is that we won't merge a one-subcommand only change, if we go through it has to be done for all commands for coherency.

let's discuss this next week as we will have had time to all review it

omar-polo:
I'd love to have something more usual than golang flags package --which is probably the only package in the stdlib that i hate with passion.  

i'd prefer to use something smaller and with less dependencies to do something as simple as parsing flags, never been a fan of fancy coloured outputs.  i'd vote for https://github.com/skeeto/optparse-go which is a small, easy, compact "port" of GNU `getopt_long` for go.

poolpOrg:
yeah but us not being fan of colored output is not representative of regular users :-)

lcheylus:
> need a bit more thinking, but one thing for sure is that we won't merge a one-subcommand only change, if we go through it has to be done for all commands for coherency.

I agree but before proposing to change all the subcommands, I preferred to submit this "small" PR only for `backup` subcommand to open discussion.

If you are OK to improve `backup` subcommand as proposed, I could also modify all sub-commands to add support for GNU-style long options.

lcheylus:
> i'd prefer to use something smaller and with less dependencies to do something as simple as parsing flags, never been a fan of fancy coloured outputs. i'd vote for https://github.com/skeeto/optparse-go which is a small, easy, compact "port" of GNU `getopt_long` for go.

1. Plakar has already fancy outputs with colors via `help` command and use of Markdown files for each subcommand.
2. https://github.com/skeeto/optparse-go Go module is too simple to parse complex flags used in subcommands. For example, `tags` flag in `backup` command need parsing (split with `,`). With https://github.com/spf13/pflag module, no need to add code to parse inputs (see https://pkg.go.dev/github.com/spf13/pflag#StringSliceVar for example).

omar-polo:
> https://github.com/skeeto/optparse-go Go module is too simple to parse complex flags used in subcommands. For example, tags flag in backup command need parsing (split with ,). With https://github.com/spf13/pflag module, no need to add code to parse inputs (see https://pkg.go.dev/github.com/spf13/pflag#StringSliceVar for example).

admittedly I dont know spf13/pflag very much, i've mentioned go-optparse because it's what I used in the past and it's nice, small (less than 300 lines IIRC and has a good license (public domain is :heart:)

I don't doubt that there could be interesting features that pflag provides -- I'll promise to take a serious look during the weekend so I can come prepared next week when we'll discuss it -- but the tags handling is a poor example, it's just a `strings.Split` away.

mathieu-plak:
Seconded... pflag is 9k5 SLOC. Admiteddly SLOC isn't a good metric but man.... golang binaries are already bloated enough.

Interesting discussion anyway, and I'm all in favor of short/longform options, the golang one just plain suck imho.

cchaudier:
Hello,

I would like to share some feedback as a Plakar user.

I have been using Plakar for a relatively short time, roughly since version **1.0.4**, and I would first like to highlight a very positive point: I really appreciate the **wording of the arguments**. The intent is clear, the vocabulary is well chosen, and it is generally easy to understand what each option is meant to do. From a semantic point of view, the CLI is well thought out.

However, from a **formal and UX perspective**, I must admit that I am sometimes a bit lost.
Today, there is a mix of arguments **with and without dashes**, and more importantly, several options that look like *long options* but are still prefixed with a **single dash**.

As a long-time Unix and Linux user — I have been using Unix/Linux systems, shells, and scripting on a daily basis for more than **25 years** — I realize that I am **losing my usual reference points**. I am used to working with a very large number of CLI tools that all follow the same implicit convention:

* `-x` for short options
* `--long-option` for long options

This convention is deeply rooted. When I see a long-named option prefixed with a single dash (`-recursive`, `-tag`, etc.), I constantly hesitate:
is this a short option? a long option? a positional argument?
This hesitation breaks the flow, especially for a tool intended to be scripted and automated.

And this is where it becomes important for me: **Plakar is a critical tool**.
I am considering using it as a **replacement for Restic** to manage backups, which means heavy usage in scripts, cron jobs, automation pipelines, and potentially in customer environments. In that context, the CLI UX is essential.

From an ergonomics standpoint, this current argument format is disturbing. It does not really match historical POSIX conventions, nor the GNU conventions that have become a **de facto standard** over the last decades. This standard is not only about style; it brings readability, predictability, and significantly reduces usage errors.

This pull request clearly **goes in the right direction**.
And I fully agree with what Gilles mentioned: if the double-dash notation is adopted, it should be **generalized to all commands**. In my opinion, this should not even be a question anymore. Today, the GNU notation (single dash for short options, double dash for long options) is widely shared and expected by Unix/Linux users.

The benefits would be immediate:

* better readability,
* easier adoption for experienced users,
* clearer scripts,
* reduced cognitive load,
* improved global consistency of the tool.

I can only encourage the Plakar team to go all the way and implement a **GNU-standard CLI notation** across the whole tool. This would significantly improve the user experience and the overall understanding of Plakar.

As a final note: I am a **cloud system administrator**, working daily with **Linux, Unix, and DevOps environments**, not only a traditional Unix/Linux admin.

One last question for the team: would you prefer that I create a dedicated **issue** for this topic to give it more visibility? This comment is more a small UX request “in a corner”, and honestly, if I had not carefully reviewed the pull requests, I would not have found this discussion. Please let me know if opening an issue would be more appropriate.

Thank you for the work done and for listening to user feedback.

poolpOrg:
I'm going to close this PR:

We have decided to tackle a rework of the CLI after v1.1.0 is out, we will take into account the feedback we received from the community, and do it globally in the entire utility.

So, stay tuned.

PR #1285, opened 2025-07-11
main: rename functions with CamelCase style

lcheylus:
Functions in this module are internal => no need to start with Uppercase.

Go naming convention for identifiers/functions:
- use CamelCase https://go.dev/ref/spec#Notation
- first char != uppercase for internal identifiers/functions https://go.dev/ref/spec#Exported_identifiers

### DIFF
--- a/main.go
+++ b/main.go
@@ -74,7 +74,7 @@ import (
 
 var ErrCantUnlock = errors.New("failed to unlock repository")
 
-func EntryPoint() int {
+func entryPoint() int {
 	// default values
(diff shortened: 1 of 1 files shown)


omar-polo (approved):
for entrypoint there's a story behind, but long story short when moving things around we forgot to lowercase it.

I'm not entirely convince about changing the names, the original ones were descriptive enough, and i'm not for blindly follow rules just because.

but since you've gone thru the effort of renaming them... ok for me.

PR #1300, opened 2025-07-16
Allow 'create' to use the passphrase from the config.

ericfaurot:
Get the passphrase from the env early and set it on the context.


brmzkw (approved):

poolpOrg (approved):
OK

PR #1302, opened 2025-07-16
basic location validation.

ericfaurot:
avoid 'plakar add foo location=blabla' errors.

### DIFF
--- a/utils/config.go
+++ b/utils/config.go
@@ -101,6 +101,14 @@ func (cl *configHandler) load(filename string, dst any) error {
 	}
 	defer f.Close()
 
+	info, err := f.Stat()
+	if err != nil {
+		return fmt.Errorf("failed to get config file info: %w", err)
+	}
+	if info.Size() == 0 {
+		return nil
+	}
+
 	err = yaml.NewDecoder(f).Decode(dst)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


brmzkw:
I don't think this is the correct fix.

The goal is to avoid creating a store with the command `plakar store add mystore location=/tmp/toto` because it creates a store at the location `location=/tmp/toto`, and not `/tmp/toto`.


The correct command is `plakar store add mystore /tmp/toto`. The location is the only parameter that should not be prefixed with it's name.

With your patch, it is impossible to create a store containing an `=` sign which might be useful. I suggest instead to only return an error if the store name begins with the extact string `location=`.

mathieu-plak:
I don't know I kinda like the idea to not let users use any dumb char available...

mathieu-plak:
And if the argument is to prevent people from using the location= syntax, I'd turn it around and accept it silently (strip the prefix) so that config is homogeneous.

poolpOrg:
It would be better to TrimPrefix(location=) because = is a valid base64 char that could be part of a bucketname,  and if we can correct unambiguously an error we should

poolpOrg (approved):

mathieu-plak (approved):

omar-polo (approved):
lgtm

PR #1303, opened 2025-07-16
do not use a go routine for emitting the report

ericfaurot:
It causes issues with logging.


poolpOrg:
as discussed, this isn't right

PR #1321, opened 2025-07-17
Rework task reporting

ericfaurot:
Tasks now create and push reports to a dedicated on a Reporter object, that processes them in a goroutine.
At start time, the scheduler create its own reporter object.
For on-demand commands, a new reporter object is created every time.


poolpOrg (approved):
OK, but as we discussed this is the first step towards moving to a journalized reporting

PR #1331, opened 2025-07-18
put more info in the manifest

ericfaurot:


plugins/plugin_loader.go
```diff
@@ -26,16 +26,23 @@ import (
...
 	Connectors []struct {
 		Type          string   `yaml:"type"`
 		Protocols     []string `yaml:"protocols"`
 		LocationFlags []string `yaml:"location_flags"`
 		Executable    string   `yaml:"executable"`
 		ExtraFiles    []string `yaml:"extra_files"`
-		Homepage      string   `yaml:"homepage"`
-		License       string   `yaml:"license"`
+		Homepage      string   `yaml:"homepage,omitempty"`
+		License       string   `yaml:"license,omitempty"`
```
omar-polo: why the omitempty?  if they're moved, we can just remove them from here.

omar-polo (approved):

plugins/plugin_loader.go
ericfaurot: right

PR #1348, opened 2025-07-23
fix(ui): always show webUI URL (#1346)

### LINKED ISSUE #1346: Plakar UI: always display the token
We should always display the following log when running `plakar ui`, regardless the value of `-no-spawn`:

```
launching webUI at http://0.0.0.0:30112?plakar_token=879e1a8f-56ea-4165-be67-6f4b74dc64c4
```

#### Context

The command `plakar ui` automatically launches a browser unless `-no-spawn` is given.

The browser is automatically redirected to `http://localhost:xx/?plakar_token=xxx`, and the UI stores the token in the local storage.

If `-no-spawn` is given, then the UI URL is displayed on stdout. It's not the case when `-no-spawn` is not given.

If the user starts another web browser to access the UI, a form appears to input the token but there's no way to get the token to enter.

s4uliu5:
The webUI URL contains a token needed to access the web UI from another browser.

Before this change, the webUI URL was only shown when using the `-no-spawn` option.

After this change, the webUI URL is always displayed.

Close #1346


poolpOrg:
the reason it was hidden is to not leak the secret on a terminal.

what is the problem you are trying to solve ?

poolpOrg:
oh I see you referenced #1346, but I think there are two things to investigate:

- why didn't it spawn a browser and if we can display only in that case
- if this could be a `-show-url` option rather than the default

brmzkw:
> the reason it was hidden is to not leak the secret on a terminal.

I am not sure it is an issue.

This is what many tools are doing. For example, in Python, werkzeug (from flask) displays a token in the terminal you should enter to run commands from a website.

Plus, `plakar ui` listens on `localhost` by default so even if the token was leaked, the impact would be limited.

poolpOrg:
discussed and i'm more worried than others

Issue #1349, opened 2025-07-23
Panic in scheduler


brmzkw:
Steps to reproduce:

1. Create a Kloset store at /var/backups: `plakar at /var/backups create -plaintext`
2. Start the agent: `plakar agent -foreground`
3. Create the following `config.yaml` file:
```yaml
agent:
  tasks:
    - name: Backup Plakar source code
      repository: /var/backups

      backup:
        path: /Users/niluje/dev/plakar/plakar
        interval: 24h
        check: true

      check:
        - interval: 1s
          path: /
          latest: true
```
4. Configure the scheduler: `plakar agent tasks configure ~/dev/plakar/config.yaml`
5. Run tasks: `plakar agent starts start`
6. Update `config.yaml`, for example set the backup interval to `5s`
7. Reconfigure: `plakar agent tasks configure ~/dev/plakar/config.yaml`

Sometimes, plakar panics with the following traceback:

```
panic: assignment to entry in nil map

goroutine 2072 [running]:
github.com/PlakarKorp/plakar/subcommands/backup.(*Backup).DoBackup(0x14000f59de8, 0x14003d9b900, 0x14000f0e0f0)
	/Users/niluje/dev/plakar/plakar/subcommands/backup/backup.go:200 +0x2e8
github.com/PlakarKorp/plakar/scheduler.(*Scheduler).backupTask(0x140004b1160, {{0x1400071b040, 0x19}, {0x14003c7a1d0, 0xc}, 0x14003d98c80, {0x14003b75b40, 0x1, 0x1}, {0x0, ...}, ...}, ...)
	/Users/niluje/dev/plakar/plakar/scheduler/tasks.go:92 +0x244
created by github.com/PlakarKorp/plakar/scheduler.(*Scheduler).Run in goroutine 2070
	/Users/niluje/dev/plakar/plakar/scheduler/scheduler.go:42 +0x238
```

Sometimes, I also have the following traceback:

```
/plakar agent -foreground
info: Plakar agent up
warn: Failed to decode RPC: failed to decode client request: EOF
warn: client write error: write unix /Users/niluje/.cache/plakar/agent.sock->: write: broken pipe
info: agent tasks configure at fs:/Users/niluje/.plakar
info: agent tasks start at fs:/Users/niluje/.plakar
info: agent tasks stop at fs:/Users/niluje/.plakar
panic: pebble: closed

goroutine 36298 [running]:
[...]

PR #1350, opened 2025-07-23
fix panic on backup in the scheduler

ericfaurot:


brmzkw:
Should close #1349

brmzkw:
I updated the issue with another traceback. I believe your patch only fixes the first one

poolpOrg (approved):

PR #1353, opened 2025-07-24
Enable demo mode with env PLAKAR_DEMO_MODE

brmzkw:
The demo mode is enabled on demo.plakar.io. With this mode, only read-only API endpoints are enabled. The UI is using the flag returned by /api/info to know whether to run in the demo mode or not, and disable the calls to the endpoints non-existing in demo mode.


poolpOrg (approved):
OK

PR #1358, opened 2025-07-24
simplify the info command and usage

poolpOrg:


omar-polo (approved):
ok but it lacks the manpage bits :p

PR #1373, opened 2025-07-27
improve diff: simplify logic and extend to work on fs.FS, makes it possible to diff local files

### LINKED ISSUE #650: Diff the content of a snapshot with the local directory
This is a request from @tusbar

It would be great to have a way to diff the content of a directory with the content of the backup, for example:

```bash
plakar backup /etc
plakar diff /etc xxxxx:/etc
```

poolpOrg:
The following PR makes diff work on fs.FS rather than VFS, making it possible to diff directories and files, including between a snapshot and local fs.

fixes #650


subcommands/diff/diff.go
```diff
@@ -43,59 +47,87 @@ func (cmd *Diff) Parse(ctx *appcontext.AppContext, args []string) error {
...
 }
 
 type Diff struct {
 	subcommands.SubcommandBase
 
-	Highlight     bool
-	SnapshotPath1 string
-	SnapshotPath2 string
+	RepositoryLocation string
+	RepositorySecret   []byte
```
omar-polo: i don't think we need this, is it a merge thing from an old branch?  aren't they in subcommand base nowadays?

omar-polo (approved):

PR #1383, opened 2025-07-28
autospawn the agent

omar-polo:
attempt to automatically spawn the agent if it's not running

even if there are windows bits in this, the windows path is not optimal.  on unix we rely on the fact that the agent will outlive the parent, as it will be reparented to init after the cli dies, on windows this doesn't work: it gets reparented at the cmd.exe level and killed afterwards.  on windows we still need a proper service.

still, it's nice =)

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -6,9 +6,11 @@ import (
 	"io"
 	"net"
 	"os"
+	"os/exec"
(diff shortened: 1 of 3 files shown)


poolpOrg (approved):
OK

PR #1401, opened 2025-07-30
Adapt to the importer/exporter/storage API changes for context.

mathieu-plak:


task/task.go
```diff
@@ -39,7 +39,12 @@ func RunCommand(ctx *appcontext.AppContext, cmd subcommands.Subcommand, repo *re
 
 	report.TaskStart(taskKind, taskName)
 	if repo != nil {
-		report.WithRepositoryName(repo.Location())
+		location, err := repo.Location()
+		if err != nil {
+			return 1, err
+		}
+
+		report.WithRepositoryName(location)
```
ericfaurot: you need to call report.TaskFailed(1, ...) before returning here, or get the location before creating the report.

ericfaurot (changes requested):

ericfaurot (approved):
ok

PR #1414, opened 2025-08-01
exporters: Adapt to new interface.

mathieu-plak:
* Exporters now need to be able to deal with Link creation.

* Implement fs and sftp while we are here as those are dead simple.


ericfaurot (approved):
ok

poolpOrg (approved):

PR #1420, opened 2025-08-01
pass parameters to plugin

poolpOrg:
this allows providing per type args in the manifest, with this it becomes possible to build importer/exporter/storage as a single executable with an argv dispatcher

### DIFF
--- a/plugins/connect.go
+++ b/plugins/connect.go
@@ -14,8 +14,8 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
-func connectPlugin(pluginPath string) (grpc.ClientConnInterface, error) {
-	fd, err := forkChild(pluginPath)
+func connectPlugin(pluginPath string, args []string) (grpc.ClientConnInterface, error) {
+	fd, err := forkChild(pluginPath, args)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


plugins/connect.go
```diff
@@ -14,8 +14,8 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
-func connectPlugin(pluginPath string) (grpc.ClientConnInterface, error) {
-	fd, err := forkChild(pluginPath)
+func connectPlugin(pluginPath string, args []string) (grpc.ClientConnInterface, error) {
```
omar-polo, line 17: the windows version needs to be sync'ed as well.

plugins/connect.go
```diff
@@ -38,15 +38,20 @@ func connectPlugin(pluginPath string) (grpc.ClientConnInterface, error) {
...
 
 	childFile := os.NewFile(uintptr(sp[0]), "child-conn")
 
-	cmd := exec.Command(pluginPath)
+	var cmd *exec.Cmd
+	if len(args) == 0 {
+		cmd = exec.Command(pluginPath)
+	} else {
+		cmd = exec.Command(pluginPath, args...)
+	}
```
omar-polo: can't you just say `args...` always?  it should work also for the case of an empty `args` slice.

plugins/connect.go
poolpOrg: resolved

plugins/connect.go
poolpOrg: resolved

mathieu-plak (approved):

ericfaurot (approved):
ok

PR #1457, opened 2025-08-06
Fix sync

brmzkw:
In 152eb08d8 we mistakenly disallowed extra args. `plakar sync` can be given a snapshot id to synchronize.


ericfaurot (approved):
ok

poolpOrg (approved):

PR #1521, opened 2025-08-07
plugins: When exec.Command pass down a context.

mathieu-plak:
* Since the plugins are basically an infinite loop, when the parent process (plakar) dies we need to command.Kill the children, which is down through context cancellation.


poolpOrg (approved):
ok

PR #1545, opened 2025-08-18
Fix version mismatch error message, closes #1444

### LINKED ISSUE #1444: Restore `plakar agent restart`, or fix error message
Now, the agent is started automatically and stopped after 1 minute by default.

However we keep this message when running plakar with a different version from the agent:

```
github.com/PlakarKorp/plakar
./plakar: agent is running with a different version of plakar (v1.0.3-devel.8c3dda2c21c9b7e4f870337462b62855a7d820a6)
To restart the agent with the current CLI version, run:
	$ plakar agent restart
```

The command `plakar agent restart` no longer exists. We should either restore it, or update this message.

brmzkw:


brmzkw:
I fear that this is not the correct patch.

Let's assume the following use-case:

1. user spawns a long-running process, for example the UI or the scheduler
2. user upgrades plakar to the latest available version
3. all commands return "version mismatch", and the only thing you can do is to either find all long-running processes and kill them, or kill the plakar agent manually

Instead, it would be great to add `plakar agent stop` back.

brmzkw:
We agreed this is not the correct fix. The current error message is not correct, but this one is too long and shouldn't advise to kill the agent.

We will likely reintroduce `plakar agent stop` and update this error message accordingly.

PR #1547, opened 2025-08-20
fix: prevent creation of empty tag on snapshots without tags

brmzkw:
Previously, when no tags were provided during backup, the snapshot was assigned a tag list with a single empty entry. Now, snapshots without tags correctly store an empty list.


poolpOrg (approved):
Yes please !

decarheil (approved):

PR #1556, opened 2025-08-23
1535 documentation for scheduler misses the start configure and stop keywords : scheduler simplification and doc is fixed

poolpOrg:


subcommands/scheduler/plakar-scheduler.1
```diff
@@ -6,10 +6,23 @@
...
 runs in the background and manages task execution based on the defined schedule.
+.Pp
+The options are as follows:
+.Bl -tag -width Ds
+.It Fl foreground
+Run the scheduler in the foreground instead of as a background service.
+.It Fl tasks Ar configfile
+Specify the configuration file that contains the task definitions and schedules.
+.It Ar stop
+Stop the currently running scheduler service.
```
omar-polo: nit: this should be `Cm stop` since it's a bare word (a command modifier), not just a placeholder for a value.

subcommands/scheduler/scheduler.go
```diff
@@ -61,11 +59,47 @@ func (cmd *Scheduler) Parse(ctx *appcontext.AppContext, args []string) error {
...
+	}
+
+	var rd io.Reader
+	if strings.HasPrefix(opt_tasks, "http://") || strings.HasPrefix(opt_tasks, "https://") {
+		resp, err := http.Get(opt_tasks)
+		if err != nil {
+			return fmt.Errorf("failed to download configuration file from %q: %w", opt_tasks, err)
+		}
+		defer resp.Body.Close()
+		rd = resp.Body
```
omar-polo: we should check the response code and maybe the content type as well.

omar-polo (approved):
OK for the improvement.

as a follow-up, we could change configureTasks to take the `scheduler.Configuration` directly and parse the config just once in `Parse`.  Maybe we can also get rid of the `ReadAll` and use the decoder to stream the reader.  Stuff for a future PR tho.

subcommands/scheduler/scheduler.go
poolpOrg: content-type i'll handle later as we still need to decide if we support multiple formats, fixed the check on status code

PR #1592, opened 2025-09-01
backup: Allow to use disk based packfiles.

mathieu-plak:
* New flag introduced for backup (only backups for now) to use disk based packfiles.

* Documentation written with much help from op@.

* depends on: https://github.com/PlakarKorp/kloset/pull/116


ericfaurot (approved):
ok

PR #1603, opened 2025-09-02
fix test: -exclude-file was renamed -ignore

omar-polo:
but while here rework the test so it actually ensure we don't backup the ignored files.


poolpOrg (approved):
OK

PR #1615, opened 2025-09-03
hide secrets unless we pass -secrets in config show

poolpOrg:


brmzkw:
Instead of having a blacklist of terms to hide, what do you think of:

1/ always display booleans and integers
2/ always hide strings, except if it's in a whitelist (location for example — which btw could contain sensitive info, so I'm not even sure it should be whitelisted)

The goal is to be safe by default.

If this discussion is refused and we decide to keep this way, note that `rclone_token` is missing in the list.

poolpOrg:
Quite frankly, I don't see how it can work the way you propose.

First of all, plakar doesn't know what keywords are used in integrations since they are now split as separate plugins projects. A few days ago, @omar-polo added a new option to s3, if we had a whitelist, that option would be obfuscated.

Second, if it was possible, it would still be considerably harder to maintain a whitelist of values to de-obfuscate than the handful of ones to obfuscate. Rclone has > 50 backends with sometimes over 10 options out of which only 1 is sensitive and contain a common pattern (like "token").

Third, if we start obfuscating non-sensitive information in a show command, people will get frustrated and will start using `-secrets` all the time, defeating the idea of providing a way to show config without disclosing the sensitive parts.

Now, for the record, I don't like my solution either as I think that the backends should provide their own config viewer and obfuscate their sensitive fields themselves, but in the meantime I think my approach is acceptable considering we had no obfuscation before and the target requires an integration interface change which we won't do right now.


> If this discussion is refused and we decide to keep this way, note that rclone_token is missing in the list.

nope, the list uses either an equalfold or a suffix match of "_"+word:

```
if strings.EqualFold(k, s) || strings.HasSuffix(k, "_"+s) {
```

so rclone_token matches the token rule because strings.HasSuffix("rclone_token, "_"+"token"):

```
$ ./plakar store show test
test:
    location: rclone://
    rclone_root_folder_id: plakar
    rclone_scope: drive
    rclone_token: '********'
    rclone_type: drive
$ ./plakar store show proton
proton:
    location: rclone://
    rclone_2fa: xxx
    rclone_password: '********'
    rclone_type: protondrive
    rclone_username: foobar
$
```

brmzkw:
ok then! thanks for the explanation, and I agree with you

mathieu-plak (approved):

PR #1619, opened 2025-09-04
Allow to force a weak passphrase

ericfaurot:
If the user enters the same unsecure passphrase twice in a row, bypass the entropy check and use it anyway.
It still has to be confirmed.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/PlakarKorp/integration-grpc v1.0.7
 	github.com/PlakarKorp/integration-ptar v1.0.0
 	github.com/PlakarKorp/integration-stdio v1.0.0
-	github.com/PlakarKorp/kloset v1.0.3-0.20250904075540-808efcbb802b
+	github.com/PlakarKorp/kloset v1.0.3-0.20250904123057-6bab4adff998
 	github.com/alecthomas/chroma v0.10.0
 	github.com/anacrolix/fuse v0.3.1
 	github.com/charmbracelet/bubbletea v1.3.6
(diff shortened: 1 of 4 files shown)


utils/utils.go
```diff
@@ -237,27 +237,46 @@ func GetPassphraseConfirm(prefix string, minEntropyBits float64) ([]byte, error)
...
+	for range retry {
 
-	passphrase2, err := readpassphrase(in, out, prefix+" passphrase (confirm): ")
-	if err != nil {
-		return nil, err
-	}
+		passphrase1, err := readpassphrase(in, out, prefix+" passphrase: ")
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%s\n", err)
+			continue
```
omar-polo: why changing this?  if readpassphrase fails, i'd keep considering it a permanent failure and stop.

utils/utils.go
```diff
@@ -237,27 +237,46 @@ func GetPassphraseConfirm(prefix string, minEntropyBits float64) ([]byte, error)
...
+			}
+		}
+		// reset our force-weak state past this point.
+		previous = nil
+		first = true
+
+		passphrase2, err := readpassphrase(in, out, prefix+" passphrase (confirm): ")
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%s\n", err)
+			continue
```
omar-polo: ditto

omar-polo (approved):
makes sense to me, leaving just two nits below.

utils/utils.go
ericfaurot: done, thanks.

utils/utils.go
ericfaurot: done

PR #1622, opened 2025-09-05
when synchronizing repositories in a single direction (to or from),

poolpOrg:
we can stop right away and warn that no snapshot was selected rather than finalize with a 0 snapshots completed.

while at it, remove references to "source" / "destiantion" in error message since we print the repo

### DIFF
--- a/subcommands/sync/sync.go
+++ b/subcommands/sync/sync.go
@@ -80,7 +80,7 @@ func (cmd *Sync) Parse(ctx *appcontext.AppContext, args []string) error {
 
 	storeConfig, err := ctx.Config.GetRepository(peerRepositoryPath)
 	if err != nil {
-		return fmt.Errorf("peer repository: %w", err)
+		return fmt.Errorf("peer store: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


poolpOrg:
before:
```
$ plakar at /tmp/a sync 1234 to /tmp/b   
info: sync: synchronization from fs:///tmp/a to fs:///tmp/b completed: 0 snapshots synchronized
$ plakar at /tmp/a sync 1234 from /tmp/b
info: sync: synchronization from fs:///tmp/a to fs:///tmp/b completed: 0 snapshots synchronized
$
```

after:
```
$ plakar at /tmp/a sync 1234 to /tmp/b  
info: No matching snapshot found in repository fs:///tmp/a
$ plakar at /tmp/a sync 1234 from /tmp/b
info: No matching snapshot found in repository fs:///tmp/b
$
```

mathieu-plak (approved):
This was not possible before, but is highly needed so yay!

brmzkw (approved):

PR #1626, opened 2025-09-05
check: improve the output

omar-polo:
* show the commit id when check fails for a snapshot;
* fix the output when check succeeds for a snapshot but a previous one failed;
* in case of an error, show the total number of snapshot that failed.


poolpOrg (approved):
OK

ericfaurot (approved):
ok

PR #1630, opened 2025-09-07
github: fix mandoc lint action

omar-polo:
the pattern might match other files, for e.g git references, so attempt to use something a bit more strict.


omar-polo:
the benefit of using got means i never noticed that find(1) could go into the repository itself... ooops =)

poolpOrg (approved):

mathieu-plak (approved):

PR #1647, opened 2025-09-09
CI: ensure manpages are installed in goreleaser.yml

brmzkw:


.github/workflows/mandoc.yml
```diff
@@ -30,3 +30,8 @@ jobs:
           cd "$dir"
           mandoc -Tlint -Wstyle -l *
           rm -rf "$dir"
+
+      - name: Ensure man pages are listed in .goreleaser.yml
+        run: |
+          diff <(find ./plakar.1 ./subcommands -name '*.1' | sort | xargs -L1 basename | sort) <(cat .goreleaser.yml | yq -r '.nfpms[0].contents[].src' | xargs -L1 basename | sort) || (echo ">> Missing man pages in .gotreleaser.yml for nfpms" && exit 1)
+          diff <(find ./plakar.1 ./subcommands -name '*.1' | sort | xargs -L1 basename | sort) <(cat .goreleaser.yml | yq -r '.homebrew_casks[].manpages[]' | xargs -L1 basename | sort) || (echo ">> Missing man pages in .gotreleaser.yml for homebrew_casks" && exit 1)
```
omar-polo: two nits:
* can we use diff -u ?  makes the output a bit more understandable when we'll forget to add something
* double sort, is that needed?

bonus: (i know i'm nitpicking quite hard -- it's just to provide a "good" example rather than annoy you), can we use find' `-print0` together with xargs -0, or at least find' -X to skip files with "funny" charecters in it?

omar-polo (approved):
nice!

.github/workflows/mandoc.yml
brmzkw: for the sort, it's needed because the man pages might be sorted in goreleaser.yml in a different order than the fs, and it shouldn't matter

For the rest, it's fixed + we now check mans [1-9], not only *.1

poolpOrg (approved):

PR #1648, opened 2025-09-09
diag: Plug a new blobsearch command.

mathieu-plak:
* This one goes through the repository packfiles directly, not through the state to search for a specific MAC. It can be slow and expensive on remotes so use with caution!

### DIFF
--- a/subcommands/diag/blob_search.go
+++ b/subcommands/diag/blob_search.go
@@ -1,0 +1,90 @@
+package diag
+
+import (
+	"encoding/hex"
+	"flag"
+	"fmt"
+	"io"
+
+	"github.com/PlakarKorp/kloset/objects"
+	"github.com/PlakarKorp/kloset/repository"
+	"github.com/PlakarKorp/kloset/repository/state"
+	"github.com/PlakarKorp/kloset/resources"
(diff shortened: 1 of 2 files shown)


subcommands/diag/blob_search.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	cmd.RepositorySecret = ctx.GetSecret()
+	cmd.ObjectID = flags.Args()[0]
+
+	return nil
+}
+
+func (cmd *DiagBlobSearch) Execute(ctx *appcontext.AppContext, repo *repository.Repository) (int, error) {
+	fmt.Fprintf(ctx.Stdout, "Warning this command is slow and expensive. Use with caution.\n")
```
omar-polo, line 38: it's in diag, do we really need this? :p

mathieu-plak: I'd rather be safe than sorry. Nowhere we say diag is dev only etc.

poolpOrg: we unplugged diag from the list of commands displayed to users, but quite frankly this is still a good reminder for us.

poolpOrg (approved):

PR #1649, opened 2025-09-09
disk based option rename

mathieu-plak:
While here also enable the option for sync.


mathieu-plak:
> lgtm; just two doubts:
> 
> * shouldn't we use `memory` (on `in-memory`) rather than `mem`?  no hard feelings on this
> * maybe, i'm still not sure, but this warrants an EXAMPLE at the end?

mem came from the big bullshit generator. No hard opinions on it honestly. 

As for the example hum yeah maybe. I find it quite simple but maybe it does warrant an example.

poolpOrg:
I agree that 'memory' is better

Issue #1653, opened 2025-09-09
Are the -log and -exclude arguments no more valid ones ?


ericxuo:
Hi,

With plakar version 1.0.2, I was using the following commands : 
plakar agent start -log $plakarLogFile
plakar at @${plakarKloset} backup -excludes /root/tmp/plakar_exclude /home

I've just run : 
go install github.com/PlakarKorp/plakar@latest
to get an updated version : 
\# plakar version
plakar/v1.0.3-devel

Now, I get the following : 
```
Starting the Plakar agent.
flag provided but not defined: -log
Usage: agent start [OPTIONS]
OPTIONS:
  -teardown duration
	    delay before tearing down the agent (default 5s)

flag provided but not defined: -excludes
Usage: backup [OPTIONS] path
       backup [OPTIONS] @LOCATION
OPTIONS:
  -check
	    check the snapshot after creating it
  -concurrency uint
	    maximum number of parallel tasks (default 7)
  -disk-based string
	    on or off or a path where to put temporary packfiles (default "off")
  -force-timestamp value
	    force a timestamp
  -ignore value
	    gitignore pattern to exclude files, can be specified multiple times to add several exclusion patterns
  -ignore-file string
	    path to a file containing newline-separated gitignore patterns, treated as -ignore
  -o value
	    specify extra importer options
  -quiet
	    suppress output
  -scan
	    do not actually perform a backup, just list the files
  -silent
	    suppress ALL output
  -tag value
	    comma-separated list of tags to apply to the snapshot
```

This means that the -log argument in the plakar agent is not allowed any more and that the -excludes argument is no more valid when creating a plakar backup.

Am I doing something wrong ?

Regards.

Xuo.

omar-polo:
Hello @ericxuo ,

`-excludes` was renamed to `-ignore-file` to avoid ambiguities (and the format now is compatible with gitignore)

Regarding `-logs` I'm not sure, I've opened #1655 to reinstate it for now, but since the agent is being retired, it's possible that we'd actually want to remove that flag.

ericxuo:
Hi, 

Does it mean that starting with Plakar 1.0.3, the agent has not to be started manually but this will be done automatically ?
So, how to write into a log file what Plakar is doing ?

Regards.

Xuo.

PR #1655, opened 2025-09-10
plakar agent start: readd some flags, improve the manpage

omar-polo:
noticed thanks to #1653, `-log` and `-foreground` were removed, but most of the logic is still there.  these changes were done in dbda028e9614e898353a619a444d472f47f0fad4, but I'm not sure about the direction: should we finish to kill the log and foreground support or readd them?

Here I'm readding them, and adjusting the documentation.  if we don't want them, given that now the agent is no longer that visible, is spawned automatically, it shuts down automatically, and we're also planning to remove it, i can amend this PR to just keep some of the manpage improvements and clean up the code.

### DIFF
--- a/subcommands/agent/plakar-agent.1
+++ b/subcommands/agent/plakar-agent.1
@@ -6,19 +6,40 @@
 .Nd Run the Plakar agent
 .Sh SYNOPSIS
 .Nm plakar agent
-.Op Cm start Fl teardown Ar delay
-.Op Cm stop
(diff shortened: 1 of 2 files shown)


omar-polo:
p.s. not re-running `sync-man.sh` because I have another PR touching the man pages.

poolpOrg (approved):
OK

mathieu-plak (approved):

PR #1666, opened 2025-09-11
set EXT=.exe when (cross)building for windows

omar-polo:


omar-polo:
example for the integration side: https://github.com/PlakarKorp/integration-s3/pull/4

would need to convert all the others now if this is ok.

PR #1676, opened 2025-09-14
Fix devel version

lcheylus:
Current version = 1.0.4-devel, not 1.0.5-devel


lcheylus:
With the official release of 1.0.4 version, this PR is no longer necessary.

PR #1686, opened 2025-09-17
move the logger implementation to plakar

ericfaurot:


poolpOrg:
close to one year old, bye bye

PR #1777, opened 2025-11-20
context: Adapt to KContext changes.

mathieu-plak:
* Cancel now expects a cause of  failure. Use nil in places where context.Canceled is fine.

* While here, when executing commands if there is an error try to read the actual underlying cause.

Depends on https://github.com/PlakarKorp/kloset/pull/179

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/PlakarKorp/integration-ptar v1.0.4
 	github.com/PlakarKorp/integration-stdio v1.0.3
 	github.com/PlakarKorp/integration-tar v1.0.0-beta.6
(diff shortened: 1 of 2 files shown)


poolpOrg (approved):
OK

main.go
```diff
@@ -400,7 +401,7 @@ func entryPoint() int {
 	go func() {
 		<-c
 		fmt.Fprintf(os.Stderr, "%s: Interrupting, it might take a while...\n", flag.CommandLine.Name())
-		ctx.Cancel()
+		ctx.Cancel(fmt.Errorf("interrupted by user"))
```
omar-polo: nit: i'd just say "interrupted" or "signal <name> received".  the user might not have (consciously) halted it.  (e.g. timeout(1) or other shenanigans).

PR #1787, opened 2025-11-27
pkg add: Check local files before fetching remote recipes

luizberti:
Previously, local .ptar files required a ./ prefix or absolute path. Now we check if the file exists locally first, falling back to remote fetch only if not found.

Forcing local resolution is still possible by passing an absolute path, and the old behavior should still work exactly the same for cases where `./` is passed, except it is now less brittle and will work with other relative paths such as `../` or `../../` and so on.

### DIFF
--- a/subcommands/pkg/plakar-pkg-add.1
+++ b/subcommands/pkg/plakar-pkg-add.1
@@ -21,8 +21,9 @@ command.
 .Pp
 Installing plugins without logging in is possible via the
 .Xr plakar-pkg-build 1
-command
-.Pq provided you have a working Go toolchain available .
+command, provided you have the necessary dependencies to build
+it locally
+.Pq currently, official plugins require make and a working Go toolchain .
 .Pp
 To force local resolution use an absolute path, otherwise to
 force remote fetching pass an HTTP or HTTPS URL.
--- a/subcommands/help/docs/plakar-pkg-add.md
+++ b/subcommands/help/docs/plakar-pkg-add.md
@@ -24,8 +24,9 @@ command.
 
 Installing plugins without logging in is possible via the
 plakar-pkg-build(1)
-command
-(provided you have a working Go toolchain available).
+command, provided you have the necessary dependencies to build
+it locally
+(currently, official plugins require make and a working Go toolchain).
 
 To force local resolution use an absolute path, otherwise to
 force remote fetching pass an HTTP or HTTPS URL.


subcommands/pkg/add.go
```diff
@@ -56,19 +63,23 @@ func (cmd *PkgAdd) Parse(ctx *appcontext.AppContext, args []string) error {
...
-			u := *baseURL
-			u.Path = path.Join(u.Path, recipe.PkgName())
-			name = u.String()
-		} else if !filepath.IsAbs(name) {
-			name = filepath.Join(ctx.CWD, name)
+		absolute := filepath.Join(ctx.CWD, name)
+		if info, err := os.Stat(absolute); err == nil && !info.IsDir() {
+			cmd.Args[i] = absolute
+			continue
 		}
```
omar-polo, line 74: this is subtly wrong. let's say name is already an absolute path, then we're mangling the it by prepending the CWD to it and then all sorts of weird things could happen.

I'd say this should be something along the lines of:

```go
absolute := name
if !filepath.IsAbs(absolute) {
        absolute = filepath.Join(ctx.CWD, absolute)
}

// ...
```

also, I'd like to explicitly check for non-regular files, not just non-directory.

subcommands/pkg/plakar-pkg-add.1
```diff
@@ -13,13 +13,19 @@ command adds a local or a remote plugin.
...
+matches an existing local file, it is installed directly.
+Otherwise, it is treated as a recipe name and downloaded from the Plakar plugin
+server which requires a login via the
 .Xr plakar-login 1
 command.
+.Pp
+Installing plugins without logging in is possible via the
+.Xr plakar-pkg-build 1
+command
+.Pq provided you have a working Go toolchain available .
```
omar-polo: nitpick: for now all the plugins are written in go, but nothing actually requires this.  The plugins only need to speak grpc, so they can be implemented in a different language.

What do you think if we say instead something along the lines of "provided that you have the build dependencies available"?

omar-polo (changes requested):
I'm leaving two minor nits that I think need to be fixed before merging this in, but otherwise it looks good to me, thank you!

subcommands/pkg/add.go
luizberti: Nice catch! I just pushed a diff for this

subcommands/pkg/plakar-pkg-add.1
luizberti: Regarding this one,

I think the message you suggested provides less useful information for people who want to try this out right now, so on top of your suggestion I also added clarification that it is only official plugins that require make and a go toolchain. Is that a good compromise?

subcommands/pkg/add.go
omar-polo: thanks!

omar-polo (approved):
Thanks!

subcommands/pkg/plakar-pkg-add.1
omar-polo: Good point!

omar-polo:
Thanks!

PR #1824, opened 2025-12-17
initial support for http mount

poolpOrg:


omar-polo (approved):
it's missing the man bits but otherwise works fine for me, very nice!

poolpOrg:
I'll update the man once I have plugged the multifs as it will change the cli

PR #1825, opened 2025-12-18
fix FUSE implementation and improve interface to allow locating specific snapshots

poolpOrg:


omar-polo (approved):
as discussed i'm still a bit confused about the inodes bookkeeping, but let's get this in and improve in tree.

PR #1838, opened 2025-12-22
do not TUI if traces are on

poolpOrg:


poolpOrg:
didn't see @brmzkw had opened #1837

PR #1844, opened 2025-12-24
get rid of stdin/out/err in context

omar-polo:
since we're agentless now, we can simplify a lot.


poolpOrg:
9 months old, no baby in sight, conflicts heavily, if we really want to get rid of stdin/out/err, it's not going to happen through this PR

PR #1853, opened 2025-12-29
one more refactor for config handling

omar-polo:
config: unify Get{Source,Store,Destination} behaviour

* let all of them to resolve internally the "at aliases" syntax
* all of them return a map or an error, not a boolean
* they verify internally that location is set

The visible changes are:
* the message for an alias not found or a location missing is
  slightly different
* passphrase_cmd gets replaced everywhere: I've checked all the
  currently packaged importers and exporters, and no-one is using it,
  I'm fine with this trade-off for now, in the near future it'll get
  replaced by a better mechanism anyway.

### DIFF
--- a/appcontext/appcontext.go
+++ b/appcontext/appcontext.go
@@ -27,6 +27,7 @@ type AppContext struct {
 func NewAppContext() *AppContext {
 	return &AppContext{
 		KContext: kcontext.NewKContext(),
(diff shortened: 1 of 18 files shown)


poolpOrg (approved):

omar-polo:
actually, i'm going to re-create this pr.

PR #1859, opened 2026-01-02
fix fuse support

poolpOrg:
this PR is a revamp of the FUSE support to fix several shortcomings

### DIFF
--- a/subcommands/mount/mount_test.go
+++ b/subcommands/mount/mount_test.go
@@ -21,7 +21,7 @@ func init() {
 	os.Setenv("TZ", "UTC")
 }
 
-func TestExecuteCmdMountDefault(t *testing.T) {
+func _TestExecuteCmdMountDefault(t *testing.T) {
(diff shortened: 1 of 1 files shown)


subcommands/mount/fuse/plakarfs/cache.go
```diff
@@ -0,0 +1,65 @@
...
+
+func newInodeCache() *inodeCache {
+	return &inodeCache{
+		files: make(map[string]*File),
+		dirs:  make(map[string]*Dir),
+	}
+}
+
+func stableKey(parts ...string) string {
+	return strings.Join(append([]string{}, parts...), "/")
```
omar-polo, line 26: nit, can't this just be `strings.Join(parts, "/")`?

poolpOrg: nope because strings.Join() wants an array

mathieu-plak (approved):

PR #1867, opened 2026-01-06
tests: Set CacheDir everywhere.

mathieu-plak:
* Now that we have local state and scanlog in sql, the Cache are not all aquired through the CacheManager, we need to also set CacheDir so that those caches are also created in that dir and garbage collected at the end of the test.

* Fixes the stray 2.0.0 dir left all over the place after running tests.


mathieu-plak:
Draft because this has a stall somewhere I'm tracking and I wanna see if the CI reproduces.

omar-polo (approved):

poolpOrg (approved):

PR #1878, opened 2026-01-12
change clone semantics, old one is problematic on multiple levels

poolpOrg:
make clone create an _empty_ store with a config cloned on the origin one.

change its UUID so there's no cache ambiguity.


poolpOrg:
discussed:

- we'll remove clone
- create will allow using another repo as a config template

PR #1903, opened 2026-01-22
New storage interface

mathieu-plak:


server/httpd/httpd.go
```diff
@@ -71,7 +71,7 @@ func (s *server) putState(w http.ResponseWriter, r *http.Request) {
 
 	var resPutIndex network.ResPutState
 	data := reqPutState.Data
-	_, err := s.store.PutState(r.Context(), reqPutState.MAC, bytes.NewBuffer(data))
+	_, err := s.store.Put(r.Context(), storage.StorageResourcePackfile, reqPutState.MAC, bytes.NewBuffer(data))
```
poolpOrg: This should be a storage.StorageResourceStateFile, not storage.StorageResourcePackfile

mathieu-plak: Nice catch fixing.

PR #1929, opened 2026-01-27
Ignore ErrNotLoggedIn on logout

aprx:
If we are already logged out, consider it as successfully failed.


poolpOrg:
that's incredible

PR #1941, opened 2026-02-05
cached: Bypass it entirely on Windows

mathieu-plak:
* Bypasses cached on Windows by issueing direct calls to RebuildState / IngestStateFile.

* As a consequence, change the way we open the repository, we need to open the cache r/w on windows.

* This fixes the windows build, and brings us back to the status-quo of previous version (no concurrent usages of plakar).


poolpOrg (approved):
was also approved by @omar-polo over the phone

PR #1976, opened 2026-02-20
fix repair crash through agent

poolpOrg:


poolpOrg:
did this with @mathieu-plak and @omar-polo

PR #1977, opened 2026-02-20
api: return an URL as location

omar-polo:
Since the switch to the new connector interface, we've been returning the storage Origin as location.  This changes the behaviour, since before we would return the original location, while now only the "hostname" part of it.

Try to reassemble the URI using the Type/Origin/Root triad instead, with some special handling for "localhost" so that fs:// paths are a bit nicer.

### DIFF
--- a/api/api_repository.go
+++ b/api/api_repository.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/PlakarKorp/kloset/connectors/storage"
+	"github.com/PlakarKorp/kloset/location"
 	"github.com/PlakarKorp/kloset/repository"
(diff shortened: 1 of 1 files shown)


api/api_repository.go
```diff
@@ -92,7 +92,14 @@ func (ui *uiserver) repositoryInfo(w http.ResponseWriter, r *http.Request) error
 		}
 	}
 
-	location := ui.repository.Origin()
+	location := ui.repository.Type() + "://"
+	if host := ui.repository.Origin(); host != "" && host != "localhost" {
```
omar-polo: use location.FLAG_LOCALFS instead of localhost check here

PR #1994, opened 2026-03-02
pkg/create: Be a bit more strict about what we accept.

### LINKED ISSUE #1935: `plakar pkg create` fails if `manifest.yaml` has a different name
Using plakar version 4de7a64d41

The command `plakar pkg create` expects two parameters:
- the `manifest.yaml` file
- a version

If we provide a file named `manifest.yml` instead, the package is successfully created but it can't be installed.

```
../plakar pkg create ./manifest.yml v1.0.0
861cdee5: OK ✓ /
861cdee5: OK ✓ /manifest.yml
861cdee5: OK ✓ /echo
Plugin created successfully: /Users/niluje/dev/plakar/plakar/toto/mys3_v1.0.0_darwin_arm64.ptar
plakar/toto on main [$?] ❯ ../plakar pkg add ./mys3_v1.0.0_darwin_arm64.ptar
[00:00] 861cdee5  ****************************************************************************************************************************************************************************************************************************** 100% 00:00     99 KiB
        export: nodes=1/0, leaves=2/2, size=99 KiB/99 KiB

✓ /
✓ echo
✓ manifest.yml






../plakar: failed to install mys3_v1.0.0_darwin_arm64.ptar: open /Users/niluje/.cache/plakar/plugins/v1.0.0/mys3_v1.0.0_darwin_arm64/manifest.yaml: no such file or directory
```

We should either force to have a file exactly named `manifest.yaml`, or, better, accept anything that matches the expected manifest format and hardcode the path.

Also the usage is not helping, it would be nice to return the same as `plakar pkg create -h`.

```
../plakar pkg create abc
../plakar: wrong usage
```

mathieu-plak:
* Check that the manifest's name is really manifest.yaml as our pkg system depends on it. We could play some tricks with the FileInfo at import time but for now this is enough.

* Fixes #1935

### DIFF
--- a/subcommands/pkg/create.go
+++ b/subcommands/pkg/create.go
@@ -27,12 +27,12 @@ import (
 	"path/filepath"
 	"runtime"
 
+	"github.com/PlakarKorp/kloset/connectors/storage"
 	"github.com/PlakarKorp/kloset/hashing"
 	"github.com/PlakarKorp/kloset/objects"
 	"github.com/PlakarKorp/kloset/repository"
 	"github.com/PlakarKorp/kloset/resources"
 	"github.com/PlakarKorp/kloset/snapshot"
-	"github.com/PlakarKorp/kloset/connectors/storage"
 	"github.com/PlakarKorp/kloset/versioning"
 	"github.com/PlakarKorp/pkg"
 	"github.com/PlakarKorp/plakar/appcontext"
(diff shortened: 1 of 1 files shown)


subcommands/pkg/create.go
```diff
@@ -80,6 +80,10 @@ func (cmd *PkgCreate) Parse(ctx *appcontext.AppContext, args []string) error {
 	cmd.Base = filepath.Dir(manifest)
 	cmd.ManifestPath = manifest
 
+	if filepath.Base(manifest) != "manifest.yaml" {
+		return fmt.Errorf("manifest's file name must be manifest.yaml")
```
omar-polo, line 84: grammar doubt moment: can we use the saxon genitive in this case?

PR #1999, opened 2026-03-04
appcontext: introduce Quiet and Silent

omar-polo:
they were previously in KContext, but it's only a client-side thing, so let's move them here.


omar-polo:
see also: https://github.com/PlakarKorp/kloset/pull/314

we can avoid bumping kloset in this case since these fields have "precedece" over the kcontext ones, and when we'll sync they're be done, so it's fine.

PR #2000, opened 2026-03-04
windows: enable the TUI; fix the terminal check

omar-polo:
term.IsTerminal(1) fails on windows because the way it's implemented on windows: it casts 1 into an handler, but it's incorrect!  It should use windows.STD_OUTPUT_HANDLE (which is -11 & (1<<32 - 1) for reasons).

so, introduce a local isTerminal() with a specific windows implementation.


omar-polo:
wait, with this I loose the cursor after a command on windows 11.  i'll have to look deply, this might need a bit more work.

omar-polo:
the terminal issues that i've seen on windows were fixed by https://github.com/PlakarKorp/plakar/pull/2067

PR #2003, opened 2026-03-05
cli: add global -json flag with JSON renderer

### LINKED ISSUE #1127: add option for json output
In order to make the life easier for api wrapper, add option (like --json) to output json as results.

would make the code https://github.com/vaultys/plakarjs here more maintainable

waveywaves:
Adds a global `-json` flag that uses a new JSON renderer to output all events as JSON lines (JSONL). This follows the existing renderer architecture — the JSON renderer sits alongside stdio and tui as a third renderer option.

Closes #1127

## Approach

Per @omar-polo's feedback, this is implemented as a **global flag with a new renderer**, not per-subcommand flags. The JSON renderer (`ui/json/`) consumes the same event stream as the stdio and tui renderers, so it works with every command automatically.

## Changes

- **`ui/json/json.go`** — New JSON renderer implementing the `ui.UI` interface. Listens to the event bus and encodes each event as a JSON line to stdout. Discards command text output (`Stdout() → io.Discard`) so only structured JSON appears.
- **`main.go`** — Adds `-json` flag and renderer selection: `-json` takes priority over `-stdio` and tui auto-detection.

## Usage

```sh
# Works with any subcommand
plakar -json backup /data
plakar -json check
plakar -json ls

# Parse in a pipeline
SNAPSHOT=$(plakar -json backup /data | jq -r 'select(.type == "result") | .snapshot')

# Each line is a JSON object with type, timestamp, snapshot, workflow, and data
{"type":"workflow.start","timestamp":"...","snapshot":"...","workflow":"backup","data":{"workflow":"backup"}}
{"type":"file.ok","timestamp":"...","snapshot":"...","workflow":"backup","data":{"path":"/data/file.txt","fileinfo":{...}}}
{"type":"result","timestamp":"...","snapshot":"...","workflow":"backup","data":{"duration":"1.23s","errors":0,"rbytes":1024,"wbytes":2048}}
```

## Design decisions

- **Global flag, not per-subcommand** — follows the renderer pattern (stdio, tui, json), works for all commands
- **JSONL format** — one JSON object per line, easy to parse with `jq`, grep, or any streaming JSON parser
- **`Stdout() → io.Discard`** — suppresses human-readable text output so only JSON events appear on stdout
- **All events emitted** — consumers can filter by `type` field (e.g., `select(.type == "result")` for summary only)

### DIFF
--- a/main.go
+++ b/main.go
@@ -30,6 +30,7 @@ import (
 	"github.com/PlakarKorp/plakar/subcommands"
 	"github.com/PlakarKorp/plakar/task"
 	"github.com/PlakarKorp/plakar/ui"
+	jsonui "github.com/PlakarKorp/plakar/ui/json"
 	"github.com/PlakarKorp/plakar/ui/stdio"
 	"github.com/PlakarKorp/plakar/ui/tui"
 	"github.com/PlakarKorp/plakar/utils"
@@ -123,6 +124,7 @@ func entryPoint() int {
 	var opt_memProfile string
 	var opt_time bool
 	var opt_trace string
+	var opt_json bool
 	var opt_stdio bool
 	var opt_quiet bool
 	var opt_silent bool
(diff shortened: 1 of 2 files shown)


omar-polo:
Hello,

Thanks for the pull request, but I think this is the wrong direction. While I'm personally not sold on having a json output, the way to implement it I believe is with the events.

We could have a global `-json` flag that is not per-subcommand, and use a different renderer: so far we have the stdio for `-stdio` or the dumb terminal, and the tui for attached terminals. Json output could use thus a new json renderer.

waveywaves:
I've reworked this to use a global -json flag with a new JSON renderer. It follows the same pattern as stdio and tui: listens to the event bus and outputs each event as a JSON line. 
Works with every subcommand automatically, no per-command changes needed as you suggested.

cc @omar-polo

main.go
omar-polo, line 175: last request :D 

@poolpOrg asked if you can swap these, `-silent` should win over `-json`.

poolpOrg: Yes, ok with the diff with these two swapped

waveywaves:
Thank you for reviewing! I have swapped the flag priority so -silent wins over -json as requested cc @omar-polo @poolpOrg

omar-polo:
Thanks!

PR #2007, opened 2026-03-07
backup: support PLAKAR_TAGS environment variable for snapshot tagging

### LINKED ISSUE #2008: Feature: PLAKAR_TAGS environment variable for snapshot tagging

In CI/CD environments (Tekton, GitHub Actions, Jenkins), tools are typically configured via environment variables rather than CLI flags. Currently, tagging snapshots requires passing `-tag` flags explicitly, which makes pipeline definitions more verbose and harder to template.

A `PLAKAR_TAGS` environment variable that allows comma-separated tags to be applied to snapshots during backup:

```sh
export PLAKAR_TAGS="ci,nightly,pipeline-run-42"
plakar backup /data
# snapshot is tagged with "ci", "nightly", "pipeline-run-42"
```

The CLI `-tag` flag should take precedence over the env var (no merging) to maintain predictable behavior.

- Using shell wrappers to inject `-tag` flags — works but adds complexity to every pipeline
- A config file approach — heavier than needed for per-run tagging
- Merging env and CLI tags — rejected for simplicity and predictability

**Additional context**

This is useful for any automation scenario where snapshot metadata should reflect the pipeline context (branch, job name, run ID) without modifying the plakar command itself.

waveywaves:
## Summary

- Adds `PLAKAR_TAGS` environment variable support to the `backup` subcommand
- Tags specified via `PLAKAR_TAGS` are comma-separated (e.g., `PLAKAR_TAGS=ci,nightly`)
- CLI `-tag` flag takes precedence over env var (no merging — if `-tag` is provided, `PLAKAR_TAGS` is ignored)
- Enables CI/CD pipelines to tag snapshots without modifying command arguments

## Motivation

In CI/CD environments (Jenkins, Tekton, GitHub Actions), it's common to configure tools via environment variables rather than flags. `PLAKAR_TAGS` lets pipelines tag snapshots declaratively:

```yaml
env:
  PLAKAR_TAGS: "ci,nightly,pipeline-run-42"
steps:
  - run: plakar backup /data
```

## Changes

- `subcommands/backup/backup.go`: Read `PLAKAR_TAGS` env var in `Parse()`, split by comma, use if no `-tag` flags provided
- `subcommands/backup/backup_test.go`: Tests for env-only, CLI-only, CLI-overrides-env, and empty env scenarios
- `subcommands/help/docs/plakar.md`: Document the new environment variable

## Test plan

- [x] `TestBackupPLAKAR_TAGS_EnvVar` — tags from env var applied to snapshot
- [x] `TestBackupCLITagOverridesPLAKAR_TAGS` — CLI `-tag` takes precedence
- [x] `TestBackupPLAKAR_TAGS_Empty` — empty env var results in no tags
- [x] `TestBackupPLAKAR_TAGS_OnlyCLI` — works normally without env var

Closes #2008

### DIFF
--- a/subcommands/backup/backup.go
+++ b/subcommands/backup/backup.go
@@ -156,6 +156,22 @@ func (cmd *Backup) Parse(ctx *appcontext.AppContext, args []string) error {
 	cmd.RepositorySecret = ctx.GetSecret()
 	cmd.Excludes = excludes
 	cmd.Tags = opt_tags.asList()
+
+	// If no tags were provided via CLI flag, check PLAKAR_TAGS env var
+	if len(cmd.Tags) == 0 {
+		if envTags, ok := os.LookupEnv("PLAKAR_TAGS"); ok && envTags != "" {
+			parts := strings.Split(envTags, ",")
+			var tags []string
+			for _, t := range parts {
+				t = strings.TrimSpace(t)
+				if t != "" {
+					tags = append(tags, t)
+				}
+			}
+			cmd.Tags = tags
+		}
+	}
+
 	cmd.Sources = flags.Args()
 
 	if len(cmd.Sources) == 0 {
--- a/subcommands/backup/backup_test.go
+++ b/subcommands/backup/backup_test.go
@@ -185,6 +185,194 @@ func TestExecuteCmdCreateWithHooks(t *testing.T) {
 	require.Contains(t, output, "backup completed")
 }
 
+func TestBackupWithPlakarTagsEnv(t *testing.T) {
+	original := os.Getenv("PLAKAR_TAGS")
+	defer os.Setenv("PLAKAR_TAGS", original)
+	os.Setenv("PLAKAR_TAGS", "daily,important")
+
+	bufOut := bytes.NewBuffer(nil)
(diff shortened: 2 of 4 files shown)


subcommands/backup/backup.go
```diff
@@ -156,6 +156,22 @@ func (cmd *Backup) Parse(ctx *appcontext.AppContext, args []string) error {
...
+			parts := strings.Split(envTags, ",")
+			var tags []string
+			for _, t := range parts {
+				t = strings.TrimSpace(t)
+				if t != "" {
+					tags = append(tags, t)
+				}
+			}
+			cmd.Tags = tags
+		}
```
omar-polo, line 172: since I'm re-reading this, small nit (which I can address after this gets merged if you prefer): the LookupEnv in this case can be replaced by just Getenv(), saving one level of indentation.

Getenv() would return an empty string if it's not set, and then we'd build a zero-length array, which is still fine.

would just save some bytes :p

poolpOrg (approved):

subcommands/backup/backup.go
poolpOrg: Lets get it in and fix ourselves, no need to delay further

PR #2028, opened 2026-03-12
assorted archive fixes

omar-polo:


omar-polo:
forgot to add: I can remove last commit if we want to keep the current atomic behaviour

PR #2031, opened 2026-03-12
Cached rework

mathieu-plak:
This fixes two outstanding issues in our implementation of cached.

1/ It fixes the zombie issues.

2/ It gets rid of the waitgroup in favor of a goroutine approach because the waitgroup semantics are not compatible with our requirements (or the other way around).

### DIFF
--- a/subcommands/cached/cached.go
+++ b/subcommands/cached/cached.go
@@ -68,6 +68,11 @@ type jobReq struct {
 	ch      chan error
 }
 
+const (
+	newJob  = 1
+	jobDone = -1
+)
+
 func (cmd *Cached) Parse(ctx *appcontext.AppContext, args []string) error {
 	var opt_foreground bool
 	var opt_logfile string
@@ -147,6 +152,14 @@ func (cmd *Cached) Execute(ctx *appcontext.AppContext, repo *repository.Reposito
 	return 0, nil
 }
 
+// Background task dealing with the teardown, basically anything running sends a
+// message over the channel, when taks is done we get another message. If the
+// teardown time passes and nothing is in flight we are free to stop, otherwise
+// someone is running and we need to stay alive. This is not a perfect "last
+// request + teardown seconds" implementation but it's close enough for our need
+// and is simpler. This is conceptually a waitgroup, except we can't use a
+// waitgroup as it has one special property (you can't reincrement the semaphore
+// while a Wait() is in progress) that our use case would transgress.
 func (cmd *Cached) Watcher(listener net.Listener) {
 	var inflight int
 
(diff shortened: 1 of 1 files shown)


subcommands/cached/cached.go
```diff
@@ -194,17 +206,15 @@ func (cmd *Cached) ListenAndServe(ctx *appcontext.AppContext) error {
 			return err
 		}
 
-		cmd.inflight.Add(1)
+		cmd.runningJobs <- 1
```
omar-polo: nit: to avoid confusions between `<- 1` vs `<- -1` (and to make it a bit more readable, I'd introduce two constants, say

```go
const (
        NewJob = 1
        JobDone = -1
) 
```

(no need to go fancy with a custom type wrapping the ints, since it's an internal thingy, just this should make it nicely readable :)

subcommands/cached/cached.go
```diff
@@ -146,6 +147,21 @@ func (cmd *Cached) Execute(ctx *appcontext.AppContext, repo *repository.Reposito
...
+	var inflight int
+
+	for {
+		select {
+		case n := <-cmd.runningJobs:
+			inflight += n
+		case <-time.After(cmd.teardown):
+			if inflight == 0 {
+				listener.Close()
+			}
```
omar-polo, line 173: (just for the archive) as discussed privately, this is not strictly equivalent to the previous behaviour, but it's a close approximation, and it's way, way simpler to reason about, so I'm all for it.

subcommands/cached/cached.go
```diff
@@ -166,19 +182,15 @@ func (cmd *Cached) ListenAndServe(ctx *appcontext.AppContext) error {
...
 	}()
 
-	go func() {
-		time.Sleep(cmd.teardown)
-		cmd.inflight.Wait()
-		listener.Close()
-	}()
-
 	for {
 		conn, err := listener.Accept()
```
omar-polo, line 208: the only problem left is that if Accept races with listener.Close and we get a new request to handle before the listener is closed, then we might hold onto the cached lock without accepting requests, and new clients will fail to run cached since the lock is already in place.

i think this could have happened before as well, in worste terms, so it's nothing too new.  but maybe we could cancel the context instead of "just" doing the listen.Close(), so that everything should fail when we're ready to close, and hopefully the client just connected will retry.

subcommands/cached/cached.go
poolpOrg: agreed

subcommands/cached/cached.go
poolpOrg: reads ok but I could use an explanation in a hackroom tomorrow, hard to grasp the reasoning by reading the code without the discussions that led to it

subcommands/cached/cached.go
mathieu-plak: Yup this is as before, and the only way to fix it is to implement retry on the other side, which we need for other work anyway. Not covered by this PR.

subcommands/cached/cached.go
mathieu-plak: done

subcommands/cached/cached.go
mathieu-plak: Tried to explain a bit in code, we can still go over this together next time.

PR #2004, opened 2026-03-05
build: add Dockerfile for CI/CD pipeline use

waveywaves:
Adds a minimal multi-stage Dockerfile and `.dockerignore` for building a container image.

Supersedes #1232 — incorporates maintainer feedback from that PR:
- **No UI-specific configuration** (per @poolpOrg and @omar-polo's feedback on #1232)
- **No docker-compose or startup scripts** — just a clean Dockerfile skeleton
- **CLI-first design** — `ENTRYPOINT ["plakar"]`, suitable for both automation (`docker run plakar backup /data`) and interactive use (`docker run -it plakar ui`)

Closes #1533

## Changes

- Multi-stage build: `golang:1.24-alpine` builder → `alpine:3.21` runtime
- `CGO_ENABLED=0` static binary (matches goreleaser config)
- Includes `ca-certificates` for HTTPS to remote repositories
- Non-root `plakar` user (`/etc/passwd` present — required by `user.Current()`)
- Preserves embedded assets (`ui/v2/frontend/`, `subcommands/help/docs/`) required by `go:embed`
- Image size: ~80MB (70.7MB binary + Alpine base)

## Usage

```sh
docker build -t plakar .
docker run --rm plakar version

# Backup in a pipeline
docker run --rm -v /data:/data plakar backup /data

# Run the UI (the image supports all subcommands)
docker run --rm -p 8080:8080 plakar ui -addr 0.0.0.0:8080
```

Issue #1533, opened 2025-08-12
Provide official docker image


CyberSinh:
**Is your feature request related to a problem?**

Many NAS operating systems use Docker as their application deployment technology (e.g., Unraid).
Installing Plakar on a NAS makes sense, especially with the HTTP backend, for setting up a backup server for machines on the local network.

**Describe the solution you'd like**

PlakarKorp should provide an official Docker image to facilitate the deployment of Plakar.

brmzkw:
Linked to #1232 

We will release a Docker image in the future, but we first need to work on the remote cache feature.

Here's a bit of context. Plakar stores an index of the Kloset store in `~/.cache/plakar`. While Plakar runs safely without this directory (it can be safely removed), having it present is important for optimal performance, as otherwise the index must be rebuilt which can be time-consuming for large stores.

We are holding off on releasing a Docker image because it might be used in serverless contexts where the cache directory is erased between runs. This results in slow starts, which could be avoided if the cache could be stored elsewhere.

Also Plakar supports hosting the Kloset store on various storage types. In #1232, the Dockerfile mounts the store in a Docker volume, which is likely not the desired approach. The Docker image we eventually release will need to support package installation, login functionality, flexible Kloset store hosting, all connectors, and more.

Given that it’s relatively easy to create a custom Dockerfile, we prefer to wait rather than provide a partially working image.

CyberSinh:
Where can I find official docker plakar image? I don't find anything on docker hub.
Thanks.

PR #2035, opened 2026-03-14
build: publish multi-arch container images to GHCR on release

waveywaves:
Builds on PR #2004 (Dockerfile) to automatically publish container images to GHCR on tagged releases.

Related: #1533

## Motivation

The Dockerfile landed in #2004, but there's no CI/CD pipeline to build and publish images. Users who want to run plakar in containers (Docker, Kubernetes, Tekton) currently have to build the image themselves.

## Changes

- **`.goreleaser.yml`** — adds `dockers_v2` configuration to publish a multi-arch image to `ghcr.io/plakarkorp/plakar` using GoReleaser's pre-built `plakar` binaries.
- **`.github/workflows/release.yml`** — adds QEMU, Docker Buildx, GHCR login, and keyless cosign signing steps. Adds `packages: write` and `id-token: write` permissions.
- **`Dockerfile`** — reuses the existing Dockerfile by adding a `goreleaser` target that copies the pre-built `$TARGETPLATFORM/plakar` binary from GoReleaser's build context. The default Dockerfile behavior still performs the full source build for manual/standalone Docker builds.

## Image tags

On a tagged release like `v1.2.0`:
- `ghcr.io/plakarkorp/plakar:v1.2.0` (multi-arch manifest)
- `ghcr.io/plakarkorp/plakar:latest` (only for stable releases, not pre-releases like `v1.2.0-beta`)

Architectures: `linux/amd64`, `linux/arm64`

## Signing

Published image tags are signed with cosign using GitHub Actions OIDC keyless signing. No private signing key is stored in the repository; the workflow receives an ephemeral certificate through the GitHub OIDC token.

Example verification:

```sh
cosign verify ghcr.io/plakarkorp/plakar:v1.2.0 \
  --certificate-identity-regexp 'https://github.com/PlakarKorp/plakar/.*' \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com
```

## Usage

```sh
docker pull ghcr.io/plakarkorp/plakar:latest
docker run ghcr.io/plakarkorp/plakar version
```

## Notes

- The `:latest` tag uses the existing `MAKE_LATEST_RELEASE` logic — only pushed and signed when the tag has no hyphen
- GoReleaser's `dockers_v2` builds the multi-arch manifest directly, so the public release tags are `:<version>` and `:latest` rather than separate architecture-suffixed tags
- The existing Dockerfile remains the source-build path for manual Docker builds; the `goreleaser` target is only for release packaging of GoReleaser's pre-built binary

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,7 @@ jobs:
       # Evaluates to the string "true" or "false". true if tag-name doesn't
       # contain a hyphen (eg. v1.2.3-dev, 1.2.4-nightly), false otherwise.
       # GoReleaser templates consume this as a string comparison for the
-      # conditional :latest container tag and cosign signing step.
+      # conditional :latest container tag.
       MAKE_LATEST_RELEASE: ${{ contains(github.ref_name, '-') == false }}
     steps:
       - name: Checkout
@@ -62,12 +62,3 @@ jobs:
           # If using goreleaser-pro, uncomment the following line
           # GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
 
-      - name: Sign container images with cosign
-        env:
-          TAG: ${{ github.ref_name }}
-          MAKE_LATEST_RELEASE: ${{ env.MAKE_LATEST_RELEASE }}
-        run: |
-          cosign sign --yes "ghcr.io/plakarkorp/plakar:${TAG}"
-          if [ "${MAKE_LATEST_RELEASE}" = "true" ]; then
-            cosign sign --yes "ghcr.io/plakarkorp/plakar:latest"
-          fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,18 +2,23 @@ name: release
 
 on:
   push:
-    # run only against tags
+    # run only against version tags
     tags:
-      - "*"
+      - "v*"
 
 permissions:
   contents: write
+  packages: write
+  id-token: write # keyless cosign signing
 
 jobs:
   goreleaser:
     runs-on: ubuntu-latest
     env:
-      # true if tag-name doesn't contain a hyphen (eg. v1.2.3-dev, 1.2.4-nightly), false otherwise
+      # Evaluates to the string "true" or "false". true if tag-name doesn't
+      # contain a hyphen (eg. v1.2.3-dev, 1.2.4-nightly), false otherwise.
+      # GoReleaser templates consume this as a string comparison for the
+      # conditional :latest container tag and cosign signing step.
       MAKE_LATEST_RELEASE: ${{ contains(github.ref_name, '-') == false }}
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -2,18 +2,23 @@ name: release
 
 on:
   push:
-    # run only against tags
+    # run only against version tags
     tags:
-      - "*"
```
waveywaves, line 7: Changed from `"*"` to `"v*"` to avoid triggering the release pipeline on non-version tags. All existing tags already use the `v` prefix (v1.0.2, v1.0.3-rc.1, v1.1.0-beta.5, etc).

waveywaves:
cc @omar-polo @mathieu-plak gentle nudge for a review

brmzkw:
Hello, and thank you for the PR.

A few questions: 
- Why do we need cosign? how does it work?
- Apparently, `dockers` from goreleaser is deprecated. Why use it and not `dockers_v2`?
- Is there any reason why not to use (and improve if necessary) the existing Dockerfile?

poolpOrg:
@waveywaves nudge in case you didn't see @brmzkw 's question

waveywaves:
Thanks @brmzkw, and thanks for the nudge @poolpOrg. Sorry I missed this.

I've pushed an update addressing the points:

- Switched from the deprecated `dockers` + `docker_manifests` config to `dockers_v2`. GoReleaser now builds the multi-arch manifest directly and publishes `ghcr.io/plakarkorp/plakar:<tag>` plus `:latest` for stable releases.
- Reused the existing `Dockerfile` instead of keeping a separate `Dockerfile.goreleaser`. The Dockerfile now has a `goreleaser` target that copies GoReleaser's pre-built `$TARGETPLATFORM/plakar` binary, while the default Dockerfile path still performs the full source build for manual/standalone Docker builds.
- Kept `cosign` intentionally: it uses GitHub Actions OIDC keyless signing (`id-token: write`), so there is no private signing key to manage. The release workflow signs the version tag, and also signs `:latest` when the release is stable. This gives users a way to verify the GHCR image provenance with `cosign verify`.

I also validated the updated GoReleaser config with:

```sh
MAKE_LATEST_RELEASE=true go run github.com/goreleaser/goreleaser/v2@latest check
```

and tested the Dockerfile's `goreleaser` target with a minimal `$TARGETPLATFORM/plakar` build context.

waveywaves:
Rebased this on latest `main`; CI is green again.

I also re-ran the GoReleaser config check locally:

```sh
MAKE_LATEST_RELEASE=true go run github.com/goreleaser/goreleaser/v2@latest check
```

Gentle nudge for another look when you get a chance.

waveywaves:
@brmzkw in a rush I didn't answer your question earlier. Cosign is required in supply chain security to cryptographically verify authenticity, ensure code integrity, and establish provenance for software artifacts. Here is exactly how it works at a glance. Signing: A developer uses a cryptographic key (or a "keyless" method) to create a digital signature for their compiled software or container image. Storage: Cosign pushes this digital signature directly to an Open Container Initiative (OCI) registry (like Docker Hub, GHCR or a private registry) right alongside the software itself. Verification: Before deploying the software, the target system uses the developer’s corresponding public key to verify that the signature matches the software artifact.

.github/workflows/release.yml
```diff
@@ -40,3 +61,13 @@ jobs:
           MAKE_LATEST_RELEASE: ${{ env.MAKE_LATEST_RELEASE }}
           # If using goreleaser-pro, uncomment the following line
           # GORELEASER_KEY: ${{ secrets.GORELEASER_KEY }}
+
+      - name: Sign container images with cosign
```
brmzkw: With the current changes:
- goreleaser pushes the docker image
- cosign is called in the github actio

Wouldn't it be better to use `docker_sign` of goreleaser? https://goreleaser.com/customization/sign/docker_sign/

waveywaves: Good call — updated in dbf79e57 to use GoReleaser `docker_signs` instead of a separate shell signing step.

The workflow still installs cosign so GoReleaser can invoke it, but signing now happens inside the GoReleaser publish phase. The `dockers_v2` tags decide whether `latest` exists, and `docker_signs` signs the generated image tags.

Validated with:

```sh
MAKE_LATEST_RELEASE=true go run github.com/goreleaser/goreleaser/v2@latest check
```

omar-polo (approved):

PR #2051, opened 2026-03-19
subcommands/mount: update usage and documentation for mount command options

dsobirk:
Current usage and documentation for mount subcommand is a bit vague. This PR clarifies options and adds a set of examples.

### DIFF
--- a/subcommands/mount/plakar-mount.1
+++ b/subcommands/mount/plakar-mount.1
@@ -6,7 +6,8 @@
 .Nd Mount Plakar snapshots as read-only filesystem
 .Sh SYNOPSIS
 .Nm plakar mount
-.Ar mountpoint
+.Op Fl to Ar mountpoint
+.Op Ar snapshotID
 .Sh DESCRIPTION
 The
 .Nm plakar mount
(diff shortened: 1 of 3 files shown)


dsobirk:
Before:
`$ plakar mount -help`
`Usage: mount PATH`

After:
`$ plakar mount -help`
`Usage: mount [-to PATH] [snapshotID]`

subcommands/mount/plakar-mount.1
```diff
@@ -17,10 +18,64 @@ This allows users to access snapshot contents as if they were part of
...
+If not specified, mount will attempt a FUSE mount in the working directory with
+a random subdirectory name.
+.It Ar snapshotID
+Optional.
+Specifies which snapshot to mount.
+If not provided, all snapshots are mounted.
+.El
 .Sh EXAMPLES
-Mount a snapshot to the specified directory:
+Mount the all snapshots to a local directory:
```
mathieu-plak: s/the//

dsobirk: I think that did it

subcommands/mount/plakar-mount.1
```diff
@@ -17,10 +18,64 @@ This allows users to access snapshot contents as if they were part of
 the local file system, providing easy browsing and retrieval of files
 without needing to explicitly restore them.
 This command may not work on all Operating Systems.
+.Pp
+In addition to the flags described below,
+.Nm plakar mount
+supports the location flags documented in
+.Xr plakar-query 7
+to precisely select snapshots.
+.Sh OPTIONS
```
omar-polo: We don't use an OPTIONS section in the manpages.

Can you please turn this into the usual sentence that we have in the SYNOPSIS

> The options are as follows:

subcommands/mount/plakar-mount.1
```diff
@@ -17,10 +18,64 @@ This allows users to access snapshot contents as if they were part of
...
+.It Fl to Ar mountpoint
+Specify the mount location.
+The
+.Ar mountpoint
+can either be:
+.Bl -bullet -offset Ds
+.It
+A directory path for FUSE mounts (e.g.,
+.Pa ~/mnt ,
+.Pa /mnt/snapshot )
```
omar-polo: I'm unsure about the `Pa` usage in this case.  `Pa` is for file system path, but it's also crucially useful for searching.  For example, with mandoc, one can do `man -k Pa=/mnt` to list all the manpages which references `/mnt` as a path.

The usage here however is just anedoctal, as neither `~/mnt` nor `/mnt/snapshot` is a well-known path meaningful to plakar.  Since this is just an example for a local path, I'm tempted to propose to don't provide an example for this.  (unlike the case below with the address, where an example is more useful.)

subcommands/mount/plakar-mount.1
```diff
@@ -17,10 +18,64 @@ This allows users to access snapshot contents as if they were part of
...
+.It
+An HTTP address including port for remote mounting (e.g.,
+.Pa http://hostname:8080 )
```
omar-polo: `Pa` here is semantically wrong, since http://... is not a file system path.  I'd either use `Dq` to (double)quote it or maybe even better with `Ql`.

subcommands/mount/plakar-mount.1
dsobirk: Done

subcommands/mount/plakar-mount.1
dsobirk: Makes sense. Since I believe that most users that would know when to do a FUSE mount would also know what a directory path is, I removed the examples.

subcommands/mount/plakar-mount.1
dsobirk: I tried with Ql

PR #2073, opened 2026-04-07
Better error message for pkg add by anonymous

brmzkw:
The previous error message was
"plakar: failed to install postgresql: authorization required" which is not obvious for new users.


mathieu-plak (approved):
Was this lost during migration to pkg package? Cause I would have sworn we had this kind of error ...

poolpOrg (approved):

PR #2074, opened 2026-04-07
Fix code preview for large files

brmzkw:
Files larger than 4096 bytes could break the ?code renderer when a token boundary happened to cross the 4096-byte chunk limit. Switch to reading the full file into memory for highlighting (up to 4MB). For larger files, disable syntax highlighting to avoid corrupted output.

Corrupted output example:
Screenshot 2026-04-07 at 18 35 19

### DIFF
--- a/api/api_snapshot.go
+++ b/api/api_snapshot.go
@@ -125,96 +125,103 @@ func (ui *uiserver) snapshotReader(w http.ResponseWriter, r *http.Request) error
 
 	switch render {
 	case "text":
-		w.Header().Set("Content-Type", "text/plain; charset=utf-8")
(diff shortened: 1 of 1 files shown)


omar-polo:
it's a pity that chroma doesn't support streaming off a reader.  since you're here, can you try instead reading the file line by line (maybe up to a maximum line size just to keep memory usage under control) and fed to the lexer one line at a time?

otherwise, your approach is probably the best we can do I think.

brmzkw:
I don't think reading line by line would work. For this JSON example it would, but it would break the coloration for multi-line strings, for example.
What do you think? Still worth trying?

mathieu-plak (approved):
What a terrible interface... but yeah okay that'll do....

poolpOrg (approved):

PR #2077, opened 2026-04-10
Update tests for Windows environment

dsobirk:
Uses windows `LocalAppData` for config and cache tests, removes windows specific parsing in `TestParseSnapshotID` since `ParseSnapshotID` explicitly adds a slash


omar-polo (approved):
Thanks!

mathieu-plak (approved):
Thx!

PR #2078, opened 2026-04-12
feat(cli): add bash completion

Toukuyokito:
Hello,

I made a completion script for plakar. I target the **release/v1.1.0-beta.7** branch because I want the most recent CLI in the projet.

Regards,

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -161,6 +161,11 @@ nfpms:
         file_info:
           mode: 0644
 
+      - src: ./completion_plakar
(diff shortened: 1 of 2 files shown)


poolpOrg:
Sorry but this is not the correct way to tackle it and is bash centric.

I'm interested in completion, but it needs to be done in code alongside implementation of commands, otherwise it will be half-baked completion just for the command name while if we do it smarter we can actually do something like:

```
$ plakar restore <TAB>
snapshot-2025-01-01
snapshot-2025-02-14

$ plakar restore snap123:/etc/<TAB>
passwd
nginx/
```

PR #2092, opened 2026-04-30
Update handling of config, cache and data directories

dsobirk:
- Add -data and -cache options similar to the existing -config options, to allow user full freedom in where to store settings and similar
- Remove directory creation from Get{Cache|Config|Data}Dir() as they are called at an early stage to calculate default locations
- Add os.MkdirAll when we have processed options

Resolves #2070

### DIFF
--- a/plakar.1
+++ b/plakar.1
@@ -8,6 +8,8 @@
 .Nm
 .Op Fl concurrency Ar number
 .Op Fl config Ar dir
+.Op Fl cache Ar dir
+.Op Fl data Ar dir
 .Op Fl cpu Ar number
 .Op Fl json
 .Op Fl keyfile Ar path
@@ -40,6 +42,14 @@ Defaults to the CPU count.
 Specify an alternate configuration directory.
 Defaults to
 .Pa ~/.config/plakar .
+.It Fl cache Ar dir
+Specify an alternate cache directory.
+Defaults to
+.Pa ~/.cache/plakar .
+.It Fl data Ar dir
+Specify an alternate data directory.
+Defaults to
+.Pa ~/.local/share/plakar .
 .It Fl cpu Ar number
 Limit the number of parallel workers
 .Nm
--- a/plakar.1
+++ b/plakar.1
@@ -7,9 +7,9 @@
 .Sh SYNOPSIS
 .Nm
 .Op Fl concurrency Ar number
-.Op Fl config Ar dir
-.Op Fl cache Ar dir
-.Op Fl data Ar dir
+.Op Fl configdir Ar dir
+.Op Fl cachedir Ar dir
+.Op Fl datadir Ar dir
 .Op Fl cpu Ar number
 .Op Fl json
 .Op Fl keyfile Ar path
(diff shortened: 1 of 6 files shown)


utils/utils.go
```diff
@@ -265,30 +265,19 @@ func GetCacheDir(appName string) (string, error) {
 
 	switch runtime.GOOS {
 	case "windows":
-		// Use %LocalAppData%
 		cacheDir = os.Getenv("LocalAppData")
-		if cacheDir == "" {
-			return "", fmt.Errorf("LocalAppData environment variable not set")
-		}
-		cacheDir = filepath.Join(cacheDir, appName)
```
omar-polo, line 273: why?  is it possible on windows for `%LocalAppData%` to not be defined?

my main concern here is that if this is empty, we end up using `~/.cache/plakar`, which on windows (I believe) would translate to `C:\Users\Foo\.cache\plakar` which seems a bit weird.

plakar.1
```diff
@@ -33,6 +33,14 @@ The following options are available:
 Specify an alternate configuration directory.
 Defaults to
 .Pa ~/.config/plakar .
+.It Fl cache Ar dir
+Specify an alternate cache directory.
+Defaults to
+.Pa ~/.cache .
```
omar-polo: isn't this `~/.cache/plakar`?

plakar.1
```diff
@@ -33,6 +33,14 @@ The following options are available:
...
+.It Fl data Ar dir
+Specify an alternate data directory.
+Defaults to
+.Pa ~/.local/share .
```
omar-polo: and this `~/.local/share/plakar`?

omar-polo (changes requested):
Hello,

Thanks for the PR!  I'm leaving some questions below, but overall it's a nice improvement, thank you!

plakar.1
dsobirk: Correct

plakar.1
dsobirk: Correct, commit amended

utils/utils.go
dsobirk: I'd agree it is very likely that `%LocalAppData%` is defined. However, if it is for whatever reason is gone, this provides a reasonable default and gives the user an opportunity to continue rather than the command refusing to run.

dsobirk: Checking one of my own systems I do actually already have a `.cache` in my `C:\Users\daniel`, together with a bunch of other .-dirs. That said, this is not the hill I will die on :-)

omar-polo: I see, and that's not a hill i want to die on either :p

That said, instead of defaulting to `.cache` on windows, could we default to `C:\Users\foo\AppData\Local` instead?  It's a tad more windows-y than `.cache` IMHO.

omar-polo (approved):
Hello :)

thanks for fixing it.  I'd have to ask you to rebase it, I've reworked a bit the manpage and now unfortunately there's a conflict :/

I was holding off a moment on this because we had to release rc1 first, now that's done and I think we can merge this.

Please consider also my comment on defaulting on windows to `~/AppData/Local` instead of `~/.cache`.

Thanks!

utils/utils.go
dsobirk: What makes most sense: \Local\, \Local\\.cache or \Local\plakar\\.cache?

dsobirk: I just went for the last option, makes the most sense to me

mathieu-plak (approved):
LGTM, Thanks a lot!

ericfaurot:
I like the diff, but I think the new options should really be named `cachedir` and `datadir`.
I also think we should introduce a new `configdir` option that overrides `config`, which we can deprecate later.

omar-polo:
Agreed, especially for `-config`, which seems to imply a file (at least in my mind) when we really need a directory containing files, `-configdir` will avoid this small confusion.  Following the same pattern, having `-datadir` and `-cachedir` with the "right" name is valuable.

but maybe we can merge this one and operate the rename in main without having to ask @dsobirk to do too many changes? :p

dsobirk:
There were multiple times that I needed to resist the temptation to call the options `datadir` etc, so I'm in favor. If you wish, I can update the flagnames, and add some kind of intelligent migration handling of the `config` flag. Won't be finished today, but should be done by Monday

omar-polo (approved):
Thanks!

PR #2098, opened 2026-05-05
improve manpages

omar-polo:
this does:
- syncs manpages with actual usages
- fixes some minor nits 
- use a consistent EXIT STATUS across manpages (DIAGNOSTICS is for error messages, not for exit status)

and also to split the command listing in plakar.1 into sections.  Please review that part carefully.  Some choices could be disagreaable.  for example, I've put "prune" under kloset commands, while "rm" is in snapshots, but this because I thought that rm is more (intuitively speaking) around snapshots, while I see prune more as a sort of GC.  I'm open to changes.


plakar.1
```diff
@@ -28,22 +32,22 @@ This can be changed either by using the
...
-Defaults to
-.Dv CPU count .
 .It Fl cpu Ar number
 Limit the number of parallel workers
 .Nm
 uses to
 .Ar number .
 By default it's the number of online CPUs.
+.It Fl json
+Use newline-delimited JSON as output format for some subcommands.
```
mathieu-plak, line 50: Why some? It was plugged as an event listener so it should turn all output into json IIRC?

mathieu-plak (commented):
Thanks for that! One nit in comment. Also currious about create-token, did it ever work ?

plakar.1
omar-polo: try `plakar -json ls`, or `info`, or...

the events are plugged only for things like backup, restore and sync (and probably the package manager by accident, since it's backup or restore), not for the others

poolpOrg (approved):

PR #2100, opened 2026-05-06
plakar policy: remove -ini

omar-polo:
it's not implemented in kloset, and it's not feasible to serialize a policy as ini since it's a nested structure.


mathieu-plak (approved):
Sure

poolpOrg (approved):

PR #2114, opened 2026-05-11
plugins: Disable idle timeout on grpc client.

mathieu-plak:
* By default grpc client uses a dialer with a 30 minutes idle timeout.

* We don't need/want this timeout to happen, for example on long lived UI instances this will break the connection for no reason. Just disable it.

* Bug found and fixed by eric@ on the enterprise side.


omar-polo (approved):

poolpOrg (approved):
OK

PR #2115, opened 2026-05-12
api: Add a NoRefresh option.

mathieu-plak:
* This option means API call won't rebuild the state. Useful one big repositories where you don't want to pay for the latency involved on each reload.


omar-polo (approved):
I wouldn't have moved the UI struct, since we're going to hopefully get rid of the Parse/Execute split in the command line soon, but why not :p

poolpOrg (approved):

PR #2117, opened 2026-05-12
kill the scheduler, it's unusable as is, cron does a better job

poolpOrg:
The scheduler in plakar is not very usable.

It was introduced too early in the project because some team members felt it was a necessity but this is not a shared view in the project and is also not a shared view in the community given how many times we were asked "why should we use it over cron ?" to which there's no good answer as objectively doesn't bring any value over using cron or whatever system timer is available.

It also didn't get as much love as other parts of the project making it lag behind and lack many of the flags and features available for most commands.

I think it should be killed and people should be encouraged to use their system cron, at least until we write a real scheduler that brings more to the table.


mathieu-plak (approved):
Bye bye

omar-polo (approved):
completely agree

mathieu-plak:
On a small note can we do something about .md gen? Yesterday when I touched a manfile I was a good citizen and only staged the corresponding md file... but I see gilles@ is not a good citizen and I can't blame him.

omar-polo:
would need to wait for a new mandoc release to be tagged, or expect ppl to build it from the last main.  I know Ingo is working on it (he needs to adapt to the lasts gnu roff releases, and from what I've heard there's quite some churn.)

PR #2119, opened 2026-05-14
don't consider "dumb" a terminal

omar-polo:
a lot of pseudo terminal emulator (emacs eshell and term, plan9port acme win, etc...) set TERM=dumb because while they act like a TTY, they don't support many (almost all?) escape codes.

Don't consider these to be a "terminal" so we automatically fall back to stdio.

### DIFF
--- a/term.go
+++ b/term.go
@@ -2,8 +2,16 @@
 
 package main
 
-import "github.com/charmbracelet/x/term"
+import (
+	"os"
+
+	"github.com/charmbracelet/x/term"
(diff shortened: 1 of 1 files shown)


omar-polo:
for reference, this is what it looks like on win in plan9port' acme the tui

image

(the second time it's with the "fixed" plakar)

poolpOrg (approved):

mathieu-plak (approved):

PR #2126, opened 2026-05-20
draft rsync subcommand

omar-polo:


poolpOrg:
superceeded by #2308 which is not quite right yet

PR #2129, opened 2026-05-23
fix(tui): fix races, ctrl+c terminal corruption and display bugs

poolpOrg:
## Summary

- **Race condition on `t.app`**: the event-loop goroutine and `Stop()` both read/wrote the `app` field without synchronization. Replaced `atomic.Bool cancel` with a `sync.Mutex` and `getApp()`/`setApp()` helpers.
- **Duplicate writes to `done` channel**: the old code had both explicit `tui.done <- …` sends inside the loop *and* a `defer close(tui.done)`, which could panic or produce unexpected `Wait()` behaviour. Replaced with a single `t.done <- result` at the very end of the goroutine.
- **ctrl+c left terminal in a bad state**: the handler used `tea.Interrupt`, which causes bubbletea to skip its normal terminal-restoration path in some versions. Switched to `tea.Quit` (which always restores the terminal) and communicate user-abort via a new `Application.aborted` flag instead of checking `tea.ErrInterrupted`.
- **`Stop()` deadlock window**: `Stop()` previously held the cancel lock while calling the blocking `app.Stop()`. A concurrent logger write trying to acquire the same lock would deadlock. Fixed by releasing the mutex before calling `app.Stop()`.
- **`switchWriter` race**: the writer is called from both the event-loop goroutine and the logger goroutine; no lock was held. Added `sync.Mutex`. Also added a `Flush()` method so partial lines are not lost at shutdown. Removed the dead `asyncTeaLogWriter` type that was never wired up.
- **`writeLastErrors` negative slice index**: `maxLines -= 3` was applied before the `<= 0` guard, so on very small terminals `maxLines` could go negative, causing a panic or wrong output. Reordered checks. Also fixed the truncation-notice condition (`start > 0`, not `maxLines < total`).
- **Events silently dropped after app error**: when `app.done` fired mid-loop, `continue` skipped the `stdio.HandleEvent` fallback for the triggering event. Now `app` is cleared so the event falls through to stdio normally.
- **Dead code removed**: unused `barPrefix` field, unused `batchMax` constant, always-nil `cmd` variable in the tick handler simplified to `return m, tick()`.

## Test plan

- [ ] Run `plakar push` on a large directory tree; verify progress bar, ETA and stats render correctly
- [ ] Press ctrl+c mid-import; verify the terminal is restored (no raw-mode residue, prompt appears on a fresh line) and plakar exits with a non-zero status
- [ ] Send SIGTERM to plakar mid-import; verify clean shutdown
- [ ] Run in a very narrow terminal (e.g. 30 columns); verify no panic from `writeLastErrors`
- [ ] Run with `--quiet` / `--json` flags; verify TUI is not activated and stdio/json renderers work normally

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/ui/tui/app.go
+++ b/ui/tui/app.go
@@ -18,13 +18,14 @@ var applications = map[string]func(*appcontext.AppContext, *Application, *reposi
 }
 
 type Application struct {
-	ctx   *appcontext.AppContext
-	job   uuid.UUID
-	name  string
-	state *State
-	done  chan struct{} // closed when Bubbletea program exits
-	prog  *tea.Program
-	err   error
+	ctx     *appcontext.AppContext
+	job     uuid.UUID
+	name    string
+	state   *State
+	done    chan struct{} // closed when Bubbletea program exits
+	prog    *tea.Program
+	err     error
+	aborted bool // true when the user pressed ctrl+c
 
 	debounceStat time.Time
 	lastStat     string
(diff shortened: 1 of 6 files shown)


poolpOrg:
THIS IS NOT MEANT TO BE COMMITTED.

PR #2130, opened 2026-05-23
mount/fuse: make the FUSE backend rock solid

poolpOrg:
## Summary

Six commits hardening the FUSE mount. Each is independently reviewable.

| # | Commit | Fix |
|---|--------|-----|
| 1 | `fix(mount/fuse): make path lookups work without prior readdir` | Direct file open returned ENOENT and certain backends hung. Root `Lookup` skipped populating the snapshot mapping (so `snapshot.Load(zeroMAC)` was called); snapshot-internal `Stat` only consulted `readDirEntries`, which is filled only by `ReadDirAll`. |
| 2 | `fix(mount/fuse): mount lifecycle, open flags, and inode key collision` | `c.MountError` check was unreachable (moved before `Serve`). `OpenDirectIO`+`OpenKeepCache` conflict (drop `DirectIO`). Snapshot inode cache key included only the 4-byte short name (collision risk). Mount with `fuse.ReadOnly()`. |
| 3 | `refactor(mount/fuse): bound the inode cache with an LRU` | Inode cache was unbounded. Now a fixed-capacity LRU (64K) with `Forget` honored. |
| 4 | `feat(mount/fuse): expose real metadata, symlinks, and xattrs` | uid/gid were the process's effective values, mode was hardcoded `0700`, no symlinks, no xattrs. Now wired from `objects.FileInfo` plus `NodeReadlinker` / `NodeGetxattrer` / `NodeListxattrer`. |
| 5 | `perf(mount/fuse): sequential-aware reads and pooled buffers` | `ReadAt` rebuilt a fresh `ObjectReader` per call, defeating its 4MB prefetch buffer. Handle now tracks next-expected offset and uses `Seek+Read` for sequential access; falls back to `ReaderAt` for random offsets. Pooled read buffers. Long attr cache TTL inside immutable snapshots. |
| 6 | `feat(mount/fuse): graceful unmount retry and optional FUSE debug` | Unmount escalates to `fusermount -u` / `diskutil unmount` before giving up. `PLAKAR_FUSE_DEBUG=1` enables request tracing. |

## Why

Triggered by two reproducible symptoms:

1. Indexers like Starfish would hang on the mount.
2. Accessing a file by absolute path (`cat /mnt/<snap>/foo/bar`) returned ENOENT unless the parent directory had been listed first.

Both shared the same root cause: the lookup path treated the readdir cache as a precondition rather than an optimization. Commit 1 fixes that; the rest are correctness, memory, fidelity, and performance issues found while in the area.

## Test plan

- [x] `go test ./subcommands/mount/...` passes locally on darwin/arm64 (unit tests on the plakarfs internals, covering the regression scenarios).
- [ ] Manual: mount a real repo, `cat` a file by direct path without prior `ls`.
- [ ] Manual: point Starfish at the mount, confirm it indexes to completion.
- [ ] Manual: `getfattr -d` on a file with xattrs, `readlink` on a symlink.
- [ ] Manual: long-running mount under load, watch RSS for growth (LRU should keep it bounded).
- [ ] Manual: SIGINT / `umount` clean-up exits without leaving a stale mountpoint.
- [ ] CI on Linux (the existing `mount_test.go` is linux-tagged and was disabled — left as-is here).

## Notes

- The existing disabled integration test in `subcommands/mount/mount_test.go` is left untouched; it's been gated behind \`_TestExecuteCmdMountDefault\` for unrelated reasons (FUSE perms in CI). New tests are unit-level on the plakarfs package and run on both linux and darwin.
- Tunables (\`rootRefresh\`, \`rootCacheTTL\`, \`kernelCacheTTL\`, cache capacity) are package constants/defaults; no CLI surface added.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/subcommands/mount/fuse/fuse.go
+++ b/subcommands/mount/fuse/fuse.go
@@ -23,7 +23,10 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
+	"os/exec"
+	"runtime"
 	"syscall"
+	"time"
 
 	"path/filepath"
 
@@ -81,13 +84,13 @@ func ExecuteFUSE(ctx *appcontext.AppContext, repo *repository.Repository, mountp
 
 	go func() {
 		<-ctx.Done()
-
-		if err := fuse.Unmount(mountpoint); err != nil {
-			fmt.Fprintf(os.Stderr, "%s is still in use; run `umount -f %s` to force\n", mountpoint, mountpoint)
-		}
+		unmountWithRetry(ctx, mountpoint)
 	}()
 
-	if err := fusefs.Serve(c, plakarfs.NewFS(ctx, repo, locateOptions, chrootfs)); err != nil {
+	server := fusefs.New(c, &fusefs.Config{
+		Debug: fuseDebugFunc(ctx),
+	})
+	if err := server.Serve(plakarfs.NewFS(ctx, repo, locateOptions, chrootfs)); err != nil {
 		return 1, err
 	}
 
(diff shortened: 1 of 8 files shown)


poolpOrg:
TEST

PR #2131, opened 2026-05-23
Add exhaustive maintenance subcommand test coverage

poolpOrg:
## Summary

Replaces the single smoke test in `subcommands/maintenance/maintenance_test.go` with 28 tests that exercise every reachable branch of the maintenance algorithm.

Statement coverage of `subcommands/maintenance/maintenance.go` goes from **47.6% → 75.2%**. The remaining 24.8% is `if err != nil` paths on kloset APIs (`HasDeletedPackfile`, `GetPackfile`, `PutLock`, `GetLocks`, `RemovePackfile`, `CommitTransaction`, `ListOrphanBlobs`, ...) reachable only with a fault-injecting storage backend, which doesn't exist in the test harness. None of those uncovered paths can reach `repository.DeletePackfile` — every one returns or `continue`s before the deletion loop at the end of `sweepPass`.

## What is covered

**Safety guards (the parts we most need to be sure of):**
- Grace period: coloured packfiles are not deleted while inside the grace window, are deleted exactly once it elapses.
- `PLAKAR_NODELETION=true`: state is updated but the bytes on disk remain readable.
- Concurrent backup: a coloured packfile resurrected via the maintenance cache before sweep is uncoloured rather than deleted (with a follow-up test pinning the current ""uncolour only persists when toDelete>0"" behaviour so future refactors are intentional).
- Lock conflicts: contended runs abort cleanly with no storage side effects; stale peer locks are removed; our own lock is released on exit.

**Colour pass:**
- Live single/multi-snapshot no-op.
- Snapshot deletion → packfiles coloured on the next run.
- Storage-only orphan inside grace ⇒ not coloured; outside grace ⇒ coloured and swept.
- Idempotent re-runs do not re-colour.

**Other branches:**
- `updateCache` deleted-snapshot cleanup via `ListDeletedSnapShots`.
- All four `PLAKAR_GRACEPERIOD` formatting cases (invalid → 7d default, 48h → 2d, 50h → 2d2h0m0s, 1h → 1h0m0s).
- `Parse` with empty args, extra args, and encrypted-repo secret propagation.
- `PLAKAR_LOCKLESS` bypass.

## Notes for the reviewer

- A few tests need an explicit `repo.RebuildState()` between maintenance calls because `CommitTransaction` writes to remote but does not merge into the in-memory aggregated state — assertions on `ListColouredPackfiles()` would otherwise miss the colourings the previous run just wrote. The pattern is encapsulated in two helpers (`primeAndDelete`, `colourRunAndRebuild`).
- One observation worth flagging from the audit: in `sweepPass`, the `CommitTransaction` is gated on `len(toDelete) > 0`. If every coloured packfile is revived by a concurrent backup in the same run, the `UncolourPackfile` calls in that pass are silently dropped. It is not unsafe — the `cache.HasPackfile` re-check in the next run still prevents deletion — but the colouring lingers in state. Both behaviours are pinned by `TestSweepUncoloursReusedPackfile` and `TestSweepUncolourPersistsWhenSomeDeletionsOccur`.

## Test plan

- [x] `go test ./subcommands/maintenance/` — 28/28 pass
- [x] `go test ./subcommands/maintenance/ -race -count=1` — clean
- [x] `go build ./...` — clean
- [x] `go vet ./subcommands/maintenance/` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/subcommands/maintenance/maintenance_test.go
+++ b/subcommands/maintenance/maintenance_test.go
@@ -44,15 +44,39 @@ func freshRepo(t *testing.T) (*repository.Repository, *appcontext.AppContext, *b
 }
 
 // runMaintenance constructs and runs a Maintenance subcommand. Output buffers
-// are cleared first so each call sees only its own log lines.
+// are cleared first so each call sees only its own log lines. Because Unlock
+// only closes the ping channel and the background goroutine performs the
+// actual DeleteLock asynchronously, we wait for the lock count to settle
+// back to its pre-Execute value before returning. Otherwise back-to-back
+// runs can race: run N+1's Lock() calls GetLocks() and sees run N's
+// not-yet-deleted lock, then races GetLock(N's MAC) against the background
+// DeleteLock and fails with "no such file or directory".
 func runMaintenance(t *testing.T, ctx *appcontext.AppContext, repo *repository.Repository,
 	bufOut, bufErr *bytes.Buffer) (int, error, string, string) {
 	t.Helper()
 	bufOut.Reset()
 	bufErr.Reset()
+
+	locksBefore, err := repo.GetLocks()
+	require.NoError(t, err)
+	preCount := len(locksBefore)
+
 	cmd := &Maintenance{}
 	require.NoError(t, cmd.Parse(ctx, nil))
 	status, err := cmd.Execute(ctx, repo)
+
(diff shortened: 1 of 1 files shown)


poolpOrg:
reviewed, all of the destructive scenarios are tested afaik

PR #2153, opened 2026-05-26
teach plakar how to build from monorepos as well

poolpOrg:


subcommands/pkg/build.go
```diff
@@ -120,12 +121,26 @@ func clone(destdir string, recipe *pkg.Recipe) error {
 		recipe.Repository = parsedUrl.String()
 	}
 
-	git := exec.Command("git", "clone", "--depth=1", "--branch", recipe.Version,
-		recipe.Repository, destdir)
+	subdir := recipe.Subdir()
+	cloneArgs := []string{"clone", "--depth=1", "--branch", recipe.Tag()}
+	if subdir != "." {
+		// For a mono-repo recipe we only need one subdirectory of the tree.
+		cloneArgs = append(cloneArgs, "--filter=blob:none", "--sparse")
```
omar-polo, line 128: why do we need this in `--depth=1` mode?

edit: referring to --filter=blob:none, i'm still wrapping my head around --sparse

omar-polo: okay, got my answer, this is to just clone and checkout only a specific subdirectory, not even the rest of the (single) tree that gets cloned.

omar-polo (approved):

subcommands/pkg/build.go
mathieu-plak: yeah I think this is more than fair otherwise it's way too heavy.

mathieu-plak (approved):
LGTM.

codecov is not happy though.

PR #2164, opened 2026-05-28
builder: Do not fail if we have a stale lock.

mathieu-plak:
* If we find a stale lock and kick it out, we need to continue the loop not fallthrough.


mathieu-plak:
@claude review

poolpOrg (approved):

PR #2175, opened 2026-06-01
pkg create: accept monorepo-esque versions

omar-polo:
in the new integrations repository we use foo/v1.2.3 as version, accept them as well.

### DIFF
--- a/subcommands/pkg/create.go
+++ b/subcommands/pkg/create.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	"github.com/PlakarKorp/kloset/connectors/storage"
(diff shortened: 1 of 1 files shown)


subcommands/pkg/create.go
```diff
@@ -68,7 +69,11 @@ func (cmd *PkgCreate) Parse(ctx *appcontext.AppContext, args []string) error {
 		version  = flags.Arg(1)
 	)
 
-	if !semver.IsValid(version) {
+	if _, v, ok := strings.Cut(version, "/"); ok {
+		version = v
+	}
+
+	if false && !semver.IsValid(version) {
```
ericfaurot: are we sure?

omar-polo: not at all! :P

PR #2179, opened 2026-06-02
backup: allow multiple ignore files

### LINKED ISSUE #1617: backup: allow more than one -ignore-file
From the users' perspective, it would make sense to separate different categories of ignore patterns. E.g., `.DS_Store` and `._*` are likely to apply to all of my macOS backups, but depending on the specific sources I want to backup, I might have additional patterns to ignore.

It would be useful to be able to call something like:
```
plakar at @mybackups backup -ignore-file macos-ignore.txt -ignore-file src-specific-ignore.txt @src
```

I could hack a script to concatenate ignore files on the fly, but who wants to live like that?

waveywaves:
Make `plakar backup -ignore-file` repeatable so users can combine common ignore files with source-specific ones.

Patterns are loaded from each `-ignore-file` in flag order, then inline `-ignore` patterns are appended as before.

Fixes #1617

Tests:
* `go test ./subcommands/backup`
* `go test ./subcommands/...`

### DIFF
--- a/subcommands/backup/backup.go
+++ b/subcommands/backup/backup.go
@@ -107,7 +107,7 @@ func (e *tagFlags) asList() []string {
 }
 
 func (cmd *Backup) Parse(ctx *appcontext.AppContext, args []string) error {
-	var opt_ignore_file string
+	var opt_ignore_files ignoreFlags
 	var opt_ignore ignoreFlags
 	var opt_tags tagFlags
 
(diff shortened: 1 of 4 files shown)


omar-polo (approved):
it looks good to me, and also a nice feature to have, thanks!

why it's still in draft?

waveywaves:
Thanks for the review! I had left it as draft while CI and a final review pass were running. Everything is green now, so I marked it ready.

poolpOrg (approved):

PR #2206, opened 2026-06-15
pkg create: add missing snap.Close()

omar-polo:


poolpOrg (dismissed):

subcommands/pkg/create.go
```diff
@@ -181,6 +181,10 @@ func (cmd *PkgCreate) Execute(ctx *appcontext.AppContext, _ *repository.Reposito
 		return 1, fmt.Errorf("failed to commit snapshot: %w", err)
 	}
 
+	if err := snap.Close(); err != nil {
+		return 1, fmt.Errorf("failed to close snapshot: %w", err)
+	}
+
```
mathieu-plak, line 187: This closes the scanCache, which is then used in CommitTransaction AFAIK to publish the delta state to teh repository. Am I missing something obvious?

mathieu-plak (changes requested):

subcommands/pkg/create.go
omar-polo: i forgot some context in this, sorry.

this diff was before i discovered the issue on windows, when i was still searching for possible causes.

I was comparing pkg/create.go with subcommands/ptar and noticed the missing close, but it wasn't the fix obviously :)

in subcommands/ptar we do close the snap, look: https://github.com/PlakarKorp/plakar/blob/main/subcommands/ptar/ptar.go#L378-L381

PR #2211, opened 2026-06-15
diag: add chunks subcommand to inspect per-file chunk layout

brmzkw:
Add `plakar diag chunks SNAPSHOT:PATH` to display the chunk breakdown of a file stored in a snapshot.

The chunks list is available from the UI, but there was no way to get them from the CLI.

Placed under `diag` since it is low-level inspection that most users won't need.

### DIFF
--- a/subcommands/diag/chunks.go
+++ b/subcommands/diag/chunks.go
@@ -1,0 +1,62 @@
+package diag
+
+import (
+	"flag"
+	"fmt"
+
+	"github.com/PlakarKorp/kloset/locate"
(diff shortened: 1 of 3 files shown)


brmzkw:
`plakar cat` checks `entry.Stat().Mode().IsRegular()`. Here I'm only checking `entry.ResolvedObject`. I'm not 100% sure which is better, but this seems to work.

poolpOrg (approved):

omar-polo (approved):

mathieu-plak (approved):

PR #2229, opened 2026-06-18
bump integrations/http ; document its store options

brmzkw:
Depends on https://github.com/PlakarKorp/integrations/pull/46/changes 

To test:

```bash
$ openssl req -x509 -newkey rsa:4096 -sha256 -days 365 -nodes -keyout server.key -out server.crt -subj "/CN=localhost"
$ ./plakar server -cert ./server.crt -key ./server.key
$ ./plakar at https://localhost:9876 ls
./plakar: failed to open the repository at https://localhost:9876: Get "https://localhost:9876/": tls: failed to verify certificate: x509: certificate signed by unknown authority
To specify an alternative repository, please use "plakar at <location> <command>".
$ ./plakar -no-tls-verify at https://localhost:9876 ls
[OK]
```

### DIFF
--- a/subcommands/config/plakar-store.1
+++ b/subcommands/config/plakar-store.1
@@ -94,6 +94,23 @@ Remove the
 for the store entry identified by
 .Ar name .
 .El
+.Ss HTTP AND HTTPS STORE OPTIONS
+When using an
+.Cm http://
+or
+.Cm https://
+location, the following additional options are available:
+.Bl -tag -width tls_no_verify
+.It Cm auth_token
+Bearer token sent in the
+.Dq Authorization
+header for every request.
+.It Cm tls_no_verify
+Set to
+.Cm true
+to disable TLS certificate verification.
+Useful when the server uses a self-signed certificate.
+.El
 .Sh EXIT STATUS
 .Ex -std
 .Sh SEE ALSO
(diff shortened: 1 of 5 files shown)


subcommands/config/plakar-store.1
```diff
@@ -94,6 +94,28 @@ Remove the
 for the store entry identified by
 .Ar name .
 .El
+.Sh HTTP AND HTTPS STORE OPTIONS
```
omar-polo: nit: i'd use `.Ss` here (sub-section) rather than a top-level section (`.Sh`, section header), because semantically it seems to me a sub-section of the DESCRIPTION rather than a top-level one.

omar-polo (commented):
there are two things i'm not sure about here:

- first `-no-tls-verify` is overly specific.  for example, on s3 we have `tls_insecure_no_verify`, and in other integrations we have different options.  maybe we should consider renaming all of them to agree on the way to spell "plz don't check tls thnx".  in any case though, what should happen on non-tls stores?  like fs?  should it fails if passed?  should be dependant on the integration?  shall we instead introduce something equivalent to `-o` but for stores? 
- second, it's nice to finally have a place where to document the options of the built-in packages.  maybe we should add the other as well.

brmzkw:
After discussion with @omar-polo , I realized this PR doesn't make sense.

For now, the only way to pass an option to a store is to use aliases. That's not perfect, but that's how it is.

Bypassing this by adding a global flag is not what we want. I'll update the PR.

omar-polo (approved):

mathieu-plak (approved):
Sure

PR #2231, opened 2026-06-19
feat: add pre-hook and post-hook in restore command

### LINKED ISSUE #2230: pre-/post- restore commands
Yes!
Following #1082 I definitely need pre-/post- restore equivalent.
Here is my use case.
UUID and GUID are different between the source of the backup and the destination of the restore. I need to change ownership. A simple way to do that is pre-/post- hooks,but unfortunately they only exist in backup command.
This happens in the context of a daily, unattended restore process

Same as #1082 but for restore :)

Playing with another systemd service that will play with ownership, but this causes async issues and adds delay between restore and post-command.

**Additional context**
N/A

Karreg:
Adds `pre-hook` and `post-hook` options for the `restore` command on the model of #1727

Fixes #2230


omar-polo (approved):
looks good to me, and extra points for adding to the manpage as well!  thanks!

poolpOrg (approved):

PR #2239, opened 2026-06-26
test: raise overall coverage to 60%

poolpOrg:
The test suite sat around 24% statement coverage. This brings it to ~61% without touching any non-test code.

Most of it is the obvious gaps: the api handlers now run against a real fs-backed repository over httptest, the subcommand packages (diag, diff, config, pkg, cached, sync, service, backup, ptar, repair, login, info, cat, check, rm, prune, dup, ui, mount/http) get their Parse and Execute paths exercised through the usual Lookup/Parse/Execute harness, and the services, server/httpd, plugins, reporting, cached, login, task and ui packages plus the root command dispatcher all gain their first real coverage.

Paths that need a live cached daemon, a real network call to api.plakar.io, a TTY for passphrase entry, or an actual OS mount are left alone — they can't run hermetically under `go test`, and forcing them would mean reshaping the code just to test it. The FUSE filesystem and the bubbletea TUI are untouched for the same reason.

(Tests largely generated with help from our friend.)

### DIFF
--- a/api/api_branches_test.go
+++ b/api/api_branches_test.go
@@ -1,0 +1,117 @@
+package api
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// VerifyMiddleware: a signature minted for one path must be rejected when the
+// request targets a different path (claims.Path mismatch).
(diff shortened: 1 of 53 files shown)


poolpOrg:
I'm an idiot and I based on the wrong branch...

PR #2262, opened 2026-07-10
Warn on snapshot backup errors during restore

### LINKED ISSUE #1821: Restore should fail if errors are reported in the snapshot
_Tested with plakar v1.0.6_

When running plakar backup /etc as an unprivileged user, some “permission denied” errors are expected, since a non-root user cannot read all files under /etc.

However, restoring the resulting snapshot completes successfully without any warnings or errors, and the restore command exits with status 0, even though some files were not backed up and therefore cannot be restored.

Should the restore command return a non-zero exit status, or at least emit warnings or errors, when the snapshot is incomplete and some files cannot be restored?

ousamabenyounes:
## What
Keep restore successful when the restore operation itself succeeds, but emit a warning when the source snapshot contains backup-time errors.

Closes #1821

## Why
The issue discussion settled on warning the user instead of making restore fail for errors that happened during backup. This makes the restore result clearer without changing the exit status for successfully restored available files.

Original issue repro on unmodified `origin/main`: create a plaintext repo, back up a tree as an unprivileged user with one unreadable file, then restore the resulting snapshot.

RED - upstream restore exits 0 and emits no restore warning even though the snapshot contains a backup error:
```text
backup_exit=0
BACKUP_STDERR_START
042ce4be: KO ✘ /tmp/src1821/denied.txt: open /tmp/src1821/denied.txt: permission denied
BACKUP_STDERR_END
snapshot=042ce4be
restore_exit=0
RESTORE_STDOUT_START
042ce4be: OK ✓ /
042ce4be: OK ✓ /readable.txt
RESTORE_STDOUT_END
RESTORE_STDERR_START
RESTORE_STDERR_END
```
Full repro log: `/tmp/see-real-bug_PlakarKorp-plakar_1821_confirmed.log`.

GREEN - patched branch warns while preserving a successful restore:
```text
ok github.com/PlakarKorp/plakar/subcommands/restore 0.211s
```

Secondary regression proof, after temporarily forcing the backup-error counter to return zero:
```text
--- FAIL: TestRestoreWarnsWhenSnapshotContainsBackupErrors (0.20s)
    restore_extra_test.go:150:
        Error: "" does not contain "snapshot contains 1 backup error"
FAIL
FAIL github.com/PlakarKorp/plakar/subcommands/restore 0.222s
```

GREEN again after restoring the fix:
```text
ok github.com/PlakarKorp/plakar/subcommands/restore 3.908s
```

## Coverage added
- Added a restore test that builds a snapshot containing a recorded backup error, restores it successfully, and asserts the warning is emitted.
- The test now corresponds to the original issue behavior: backup-time errors are visible in the snapshot, but restore previously completed silently.

## Test plan
- `docker run --rm -v "$PWD":/src -v /tmp/plakar-go-cache:/go/pkg/mod -v /tmp/plakar-go-build:/root/.cache/go-build -w /src golang:1.25 go test ./subcommands/restore -run TestRestoreWarnsWhenSnapshotContainsBackupErrors -count=1`
- `docker run --rm -v "$PWD":/src -v /tmp/plakar-go-cache:/go/pkg/mod -v /tmp/plakar-go-build:/root/.cache/go-build -w /src golang:1.25 go test ./subcommands/restore -count=1`
- `docker run --rm -v "$PWD":/src -v /tmp/plakar-go-cache:/go/pkg/mod -v /tmp/plakar-go-build:/root/.cache/go-build -w /src golang:1.25 sh -c 'go build -v ./... && go test ./...'`

Full validation result:
```text
exit:0
go build -v ./...: pass
go test ./...: pass
```

### DIFF
--- a/subcommands/restore/restore.go
+++ b/subcommands/restore/restore.go
@@ -200,7 +200,34 @@ func (cmd *Restore) Execute(ctx *appcontext.AppContext, repo *repository.Reposit
 			return 1, err
 		}
 
+		backupErrorCount, err := snapshotBackupErrorCount(snap)
+		if err != nil {
+			ctx.GetLogger().Warn("could not inspect snapshot backup errors: %s", err)
+		} else if backupErrorCount != 0 {
+			errorWord := "errors"
+			if backupErrorCount == 1 {
+				errorWord = "error"
+			}
+			ctx.GetLogger().Warn("snapshot contains %d backup %s; restored files are limited to data that was backed up", backupErrorCount, errorWord)
+		}
+
 		snap.Close()
 	}
 	return 0, nil
+}
+
+func snapshotBackupErrorCount(snap *snapshot.Snapshot) (int, error) {
+	fsc, err := snap.Filesystem()
+	if err != nil {
+		return 0, err
+	}
+
+	count := 0
+	for _, err := range fsc.Errors("/") {
+		if err != nil {
+			return 0, err
+		}
+		count++
+	}
+	return count, nil
 }
--- a/main.go
+++ b/main.go
@@ -88,6 +88,36 @@ func progName() string {
 	return filepath.Base(os.Args[0])
 }
 
+const fallbackUsername = "unknown"
+
+func currentUserDefault(cwd string) *user.User {
+	userDefault, err := user.Current()
+	if err != nil {
+		username := os.Getenv("USER")
+		if username == "" {
+			username = os.Getenv("LOGNAME")
+		}
+		if username == "" {
+			username = fallbackUsername
+		}
+
+		userDefault = &user.User{
+			Username: username,
+		}
+	}
+
+	homeDir := os.Getenv("HOME")
+	if homeDir == "" {
+		homeDir = userDefault.HomeDir
+	}
+	if homeDir == "" {
+		homeDir = cwd
+	}
+	userDefault.HomeDir = homeDir
+
+	return userDefault
+}
+
 func entryPoint() int {
 	// default values
 	cwd, err := os.Getwd()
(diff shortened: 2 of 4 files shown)


subcommands/restore/restore.go
```diff
@@ -50,6 +50,15 @@ type Restore struct {
...
 }
 
+const (
+	snapshotErrorsRootPath       = "/"
+	backupErrorsInspectWarnFmt   = "could not inspect snapshot backup errors: %s"
+	backupErrorsRestoreWarnFmt   = "snapshot contains %d backup %s; restored files are limited to data that was backed up"
+	backupErrorSingular          = "error"
+	backupErrorPlural            = "errors"
+	backupErrorSingularThreshold = 1
+)
```
omar-polo: these are a bit too much.  they're used basically only once, so it's fine to have them inline.  all these constants make the code actually harder to read IMHO.

omar-polo (changes requested):

mathieu-plak:
This PR is a nice idea, but we have conflicting views in the team on the behavior of restore. This is something we need the whole team to agree on (and it might even be a kloset knob) before proceeding, so this PR might stay in draft for a while before we reach a consensus (in which case we will let you know!)

mathieu-plak:
Oh and also please can you drop the "semantic" commit messages? As you can see we don't use them in this repo so it could be cool if we kept the same convention!

ousamabenyounes:
Thanks @omar-polo and @mathieu-plak. I addressed the actionable cleanup here:

- removed the one-off restore warning constants and inlined those values where they are used;
- renamed the commit and PR title to avoid the semantic `fix(scope): ...` style.

Local validation:
- `go test ./subcommands/restore -count=1 -timeout 2m -v` passed;
- `go test ./... -count=1 -timeout 5m` passed.

I understand the behavior itself still needs team consensus, so this can stay pending on that decision.

ousamabenyounes:
And regarding the semantic commit messages, I’m adding this convention to my workflow so that my future PRs follow YOUR standards. :)

ousamabenyounes:
Rebased on current main; no behavior changes beyond the existing PR diff.\n\nLocal validation after rebase:\n- go test ./subcommands/restore -count=1\n- go build -v ./...\n\nThe remaining blocker still looks like the team consensus noted above, not an unresolved code thread.

PR #2279, opened 2026-07-22
ci: actually propagate tests failure

omar-polo:
`run` is executed in an environment where `errexit` is set, so if the tests fail, we don't get the chance to execute the rest of the script.

while here, catch the failure also if `go install` fails.

a port of https://github.com/PlakarKorp/kloset/pull/493


mathieu-plak (commented):
yup

mathieu-plak (approved):

PR #2287, opened 2026-07-25
Surface packfile temp creation failures during backup

### LINKED ISSUE #2257: [Bug]: import: aborted when /tmp is full
### Describe the bug

Hello

When /tmp is full the backup is failing with this message : 

ubuntu@flop-test:~$ plakar-1.1.4 at @ovh-store backup korpus/
[00:00] import: aborted !

### Steps to reproduce

1- Have /tmp full
2- run a backup from the fs

### Expected behavior

_No response_

### Actual behavior

```shell

```

### Plakar version

_No response_

### Plugin versions

```shell

```

### OS, version and architecture

_No response_

### Store location

_No response_

### Backup source location

_No response_

### Restore destination (if applicable)

_No response_

### Debug output

```shell

```

### Additional context

_No response_

ousamabenyounes:
## Summary

Fix #2257 by surfacing the real packfile temporary-file creation failure instead of masking it behind `context.Canceled`.

This keeps the default packfile temporary location unchanged (`os.MkdirTemp("", ...)`, normally `/tmp`). When kloset cancels the context with a concrete cause, backup now returns that cause from `snap.Backup()` or `snap.Commit()` failures. The helper lives on `AppContext` so other commands can reuse the same cancellation-cause handling.

## Validation

RED -> GREEN proof:

- RED: with the production helper removed, `go test ./appcontext ./subcommands/backup -run "TestErrorCause|TestBackupPropagatesContextCause" -count=1 -timeout 2m -v` fails because `ctx.ErrorCause` is undefined.
- GREEN: the same targeted command passes with the fix.

Additional local validation:

- `./run-ci.sh` passes in Docker `golang:1.25`.
- `origin/main` full baseline also passes in Docker `golang:1.25`.

### DIFF
--- a/subcommands/backup/backup.go
+++ b/subcommands/backup/backup.go
@@ -392,6 +392,7 @@ func (cmd *Backup) DoBackup(ctx *appcontext.AppContext, repo *repository.Reposit
 		}
 
 		if err := snap.Backup(source); err != nil {
+			err = ctx.ErrorCause(err)
 			if err := executeHook(ctx, cmd.FailHook); err != nil {
 				ctx.GetLogger().Warn("post-backup fail hook failed: %s", err)
 			}
@@ -400,6 +401,7 @@ func (cmd *Backup) DoBackup(ctx *appcontext.AppContext, repo *repository.Reposit
 	}
 
 	if err := snap.Commit(); err != nil {
+		err = ctx.ErrorCause(err)
 		if err := executeHook(ctx, cmd.FailHook); err != nil {
 			ctx.GetLogger().Warn("post-backup fail hook failed: %s", err)
 		}
--- a/appcontext/appcontext.go
+++ b/appcontext/appcontext.go
@@ -1,6 +1,9 @@
 package appcontext
 
 import (
+	"context"
+	"errors"
+
 	"github.com/PlakarKorp/kloset/connectors"
 	"github.com/PlakarKorp/kloset/kcontext"
 	"github.com/PlakarKorp/pkg"
@@ -43,6 +46,15 @@ func NewAppContextFrom(ctx *AppContext) *AppContext {
 // XXX: This needs to go away progressively by migrating to AppContext.
 func (c *AppContext) GetInner() *kcontext.KContext {
 	return c.KContext
+}
+
+func (c *AppContext) ErrorCause(err error) error {
+	if errors.Is(err, context.Canceled) {
+		if cause := context.Cause(c.Context); cause != nil {
+			return cause
+		}
+	}
+	return err
 }
 
 func (c *AppContext) SetSecret(secret []byte) {
--- a/appcontext/appcontext_test.go
+++ b/appcontext/appcontext_test.go
@@ -2,6 +2,8 @@ package appcontext
 
 import (
 	"bytes"
+	"context"
+	"errors"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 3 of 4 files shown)


omar-polo (commented):
I'm not sure this is the right approach.  The storage for the packfiles is already customizable, and CacheDir is not necessarely better than /tmp.

The real issue here is there we're not bubbling up correctly one error from kloset, not the location of the packfiles.

ousamabenyounes:
Thanks for the review @omar-polo. Addressed it in a9702bee: the default packfile temp location is unchanged now, and the fix is focused on surfacing the real temporary packfile creation failure instead of hiding it behind cancellation/aborted backup output.

Validation:

- RED: `TestBackupPropagatesPackfileCreationError` timed out after 20s with the production fix removed, while the kloset packer error was not surfaced.
- GREEN: the focused test passes after the fix.
- `go test ./subcommands/backup -coverprofile=backup.coverage.out -covermode=atomic -count=1 -timeout 2m -v` passes with 72.3% coverage.
- Diff coverage is 100% on changed production lines.
- Local full `run-ci.sh` is iso-baseline: `origin/main` and this branch both fail the root package with `plakar: go away casper !`; no new backup failure from this branch.
- GitHub checks are green: both `build` jobs plus Codecov patch/project passed.

subcommands/backup/backup.go
```diff
@@ -455,6 +471,31 @@ func (cmd *Backup) DoBackup(ctx *appcontext.AppContext, repo *repository.Reposit
 	return 0, nil, snap.Header.Identifier, warning
 }
 
+func validatePackfileTempStorage(dir string) error {
```
mathieu-plak: I'm unsure on what this buys us?

subcommands/backup/backup.go
```diff
@@ -455,6 +471,31 @@ func (cmd *Backup) DoBackup(ctx *appcontext.AppContext, repo *repository.Reposit
...
+	}
+	if err := os.Remove(name); err != nil {
+		return fmt.Errorf("failed to remove temporary packfile: %w", err)
+	}
+	return nil
+}
+
+func backupErrorCause(ctx *appcontext.AppContext, err error) error {
+	if errors.Is(err, context.Canceled) {
+		if cause := context.Cause(ctx.Context); cause != nil {
```
mathieu-plak: Indeed, we added Cancellation on the packer managers as a mean of aborting from a detached background job. But we never actually used the Cause that's why it's hidden! Nice one. The only thing is that I think this belongs on AppContext, because we will have other places that needs it (synchronize comes to mind).

subcommands/backup/backup.go
```diff
@@ -83,6 +85,13 @@ func (e *ignoreFlags) Set(value string) error {
 
 type tagFlags string
 
+const (
+	packfileTempStorageMemory = "memory"
+	packfileTempDirPrefix     = "plakar-backup-"
+	packfileTempDirSuffix     = "-*"
+	packfilePreflightPrefix   = "plakar-pack-"
+)
```
mathieu-plak: Can you avoid doing those? I know  const are a good thing, but we are abusing it here a bit (packfileTempStorageMemory for example is used literally once). If you feel like you really need to do it then at least do it in a separate commit please, this makes the diff harder to read than need be.

mathieu-plak (changes requested):
A few comments to address inline! 
One meta comment, please rebase that PR to drop the first commit it is now useless and we don't want it in the history.

Thanks for your patience, keep up the good work!

subcommands/backup/backup.go
omar-polo: yeah.. it smells of TOCTOU, i'd just drop it.

ousamabenyounes:
Thanks for the reviews. Addressed the requested rework in 6d7f6567:

- dropped the TOCTOU preflight check and kept the default packfile temp behavior unchanged;
- moved cancellation-cause unwrapping to `AppContext`;
- removed the one-off constants and squashed the branch to one commit.

Validation: reproduced the missing helper as RED, then reran the targeted tests and full local `./run-ci.sh` GREEN in Docker `golang:1.25`.

mathieu-plak (approved):
Lovely!

poolpOrg (approved):
LGTM

Issue #2254, opened 2026-07-01
[Bug]: Plakar backup fails when cache drive is full


dlarochette:
### Describe the bug

A backup task failing while running leaves the cache full
A backup task fails when the cache is full from the start
Nowhere, plakar seems to be able to clean its own cache, except at the end of a successful backup?!

In my case, I have 1 2T backup and one 1.2T backup plus various others.

### Steps to reproduce

Run a task with massive cache usage, then run any following task

### Expected behavior

Whenever a backup starts, or in the middle of a backup task, if the cache drive is full, plakar should initiate a cleanup of the cache, according to multiple running backup/restore tasks, maybe privileging quick-to-end tasks so that their cache cleaned leaving space for others.

### Actual behavior

The current backup fails, leaving the cache full.
All the following backup tasks fail, also leaving cache full

### Plakar version

```shell
$ plakar version                                        
plakar/v1.1.3
```

### Plugin versions

```shell
$ plakar pkg show
mysql@v1.1.0-beta.4
postgresql@v1.1.0-beta.8
s3@v1.1.2
sftp@v1.1.3
```

### OS, version and architecture

```shell
$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"
VERSION_CODENAME=trixie
DEBIAN_VERSION_FULL=13.5
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

### Store location

```shell
$ df -h /home/save/.cache/plakar
Filesystem                     Size  Used Avail Use% Mounted on
/dev/mapper/vg01-plakar_cache   49G  488M   46G   2% /home/save/.cache/plakar
```

### Backup configuration

```ini
# /etc/systemd/system/backup-[redacted]-db.service
[Unit]
Description=Sauvegarde quotidienne de la base de données [redacted] sur [redacted]
After=network.target
Before=plakar_prune-[redacted].service

[Service]
Type=oneshot
LoadCredentialEncrypted=save_id_ed25519
LoadCredentialEncrypted=[redacted].kloset.passphrase
User=save
Group=save
[...]

PR #2288, opened 2026-07-25
backup: clean partial scan cache on create failure

ousamabenyounes:
## Summary
- generate the backup snapshot ID before calling `snapshot.Create` so Plakar can identify the scan cache path for that attempt
- remove only that partial scan cache directory if snapshot creation fails before returning a builder
- add a regression test covering a scan cache constructor that leaves a partial directory before failing

## Testing
- RED before fix: `go test ./subcommands/backup -run TestBackupCleansPartialScanCacheOnCreateFailure -count=1` failed because the partial scan cache directory remained
- GREEN after fix: `go test ./subcommands/backup -run TestBackupCleansPartialScanCacheOnCreateFailure -count=1`
- `go test ./subcommands/backup -count=1`
- `go test ./subcommands/pkg -count=1 && go test ./subcommands/backup -count=1`
- local full runner on patched branch: `go build -v ./...`, `go test -v -p 4 -coverprofile=coverage.out -covermode=atomic -timeout 2m ./...`, `GOOS=windows GOARCH=amd64 go build -v .`

Refs #2254

### DIFF
--- a/subcommands/backup/backup.go
+++ b/subcommands/backup/backup.go
@@ -68,20 +68,20 @@ type Backup struct {
 	Perimeter           string
 }
 
-const backupScanCacheName = "scan"
+const snapshotWorkCacheName = "scan"
 
 func init() {
 	subcommands.Register(func() subcommands.Subcommand { return &Backup{} }, 0, "backup")
 }
 
-func cleanupPartialScanCache(ctx *appcontext.AppContext, identifier objects.MAC) {
+func cleanupPartialSnapshotWorkspace(ctx *appcontext.AppContext, identifier objects.MAC) {
 	if ctx.CacheDir == "" || identifier == objects.NilMac {
 		return
 	}
 
-	scanCacheDir := filepath.Join(ctx.CacheDir, caching.CACHE_VERSION, backupScanCacheName, fmt.Sprintf("%x", identifier))
-	if err := os.RemoveAll(scanCacheDir); err != nil {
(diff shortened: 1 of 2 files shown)


omar-polo (commented):
I'm not sure, but aren't we already removing the packfile directory a few lines above (see [backup.go:298](https://github.com/PlakarKorp/plakar/blob/main/subcommands/backup/backup.go#L298)) so how is this changing anything?

(also, it's not really a scan cache, that's something we don't really have anymore, but rather the storage for packfiles that are in the process of being created)

ousamabenyounes:
Sorry for the delayed reply, and thanks for the review.

You're right that this PR doesn't fix #2254 as reported. Line 298 removes
`PackfileTempStorage`, while this PR removes what kloset still calls the
ScanCache — happy to use whatever term you prefer for it.

The current Plakar workaround duplicates kloset's internal path and only cleans
the new failed attempt. It won't recover the large workspace orphaned by an
earlier interrupted backup.

I think this needs two separate changes:

1. Error-path cleanup in kloset: `Create`, `CreateWithRepositoryWriter` and
   `Fork` can return after `newBuilder` without closing the ScanCache, so
   `DeleteOnClose` never fires (checked against 9310b71, the revision in go.mod).
   Small and self-contained, but it only covers empty workspaces.
2. Stale-workspace recovery for #2254 itself, which has to stay safe with
   concurrent backups. That's a design question, not a patch.

I'll close this PR unless you'd prefer to keep it as limited hardening. Happy to
send the kloset error-path PR straight away; for the recovery side I'd rather
agree on the shape first.

PR #2290, opened 2026-07-26
pkg: remove mkstruct()

omar-polo:
it's a relic of the ancient times when dragons crossed the skys and importers were required to provide a full VFS, nowadays kloset fills the gaps in it.


mathieu-plak (approved):
What a bad name this was :'

PR #2292, opened 2026-07-28
pkg/create: Simplify the ptar creation.

mathieu-plak:
* While hunting the CI bug I stumbled upon this one. The ptar creation for package doesn't need to use the low level multi snapshot API of the builder. It can just use Create/Backup and eveyrthing is handled for it.

* I think (I'm sure even) this was copied over from ptar, which needs this because it supports encapsulating multiple snapshots.

* No behavior change.


omar-polo (approved):
lovely!

PR #2293, opened 2026-07-28
api: consolidate coverage

omar-polo:
dedup, reorganize regress.  while here, remove some duplicated ones as well.


mathieu-plak (approved):
Love it.

PR #2295, opened 2026-07-31
ls: report the errors recorded in a snapshot

### LINKED ISSUE #1268: List errors in `plakar ls`
Related to #1255 

The command `plakar ls` lists files on stdout. We want to also list the errors on stderr.

Expected output:

```
2025-04-12T05:16:09Z -rw-r--r--     root    wheel    515 B afpovertcp.cfg
2025-04-12T05:16:09Z Lrwxr-xr-x     root    wheel     15 B aliases -> postfix/aliases
aliases.db: xattr.list: permission denied
2025-04-12T05:16:09Z drwxr-xr-x     root    wheel    288 B apache2
2025-04-12T05:16:09Z drwxr-xr-x     root    wheel    512 B asl
2025-04-12T05:16:09Z -rw-r--r--     root    wheel   1.1 kB asl.conf
2025-04-12T05:16:09Z -rw-r--r--     root    wheel    149 B auto_home
2025-04-12T05:16:09Z -rw-r--r--     root    wheel    195 B auto_master
2025-04-12T05:16:09Z -rw-r--r--     root    wheel   1.9 kB autofs.conf
2025-04-12T05:16:09Z -r--r--r--     root    wheel    265 B bashrc
```

When done, we should also remove the subcommand `plakar diag errors`.

0xfandom:
`plakar ls SNAPSHOT:PATH` lists the entries of a snapshot, but files that could not be read when the backup ran are simply absent from it, with nothing saying why. The errors recorded for them were only reachable through a separate command, `plakar info -errors`.

This lists them on stderr, merged into the listing on stdout.

`vfs.Errors(beneath)` yields items sorted by pathname, so they merge into the `WalkDir` output directly: before printing an entry, drain every error that sorts before it, then drain what is left once the walk is done.

Errors below the listed directory are only reported for a recursive listing, matching the entries themselves. Names follow the same convention as the listing too: base name for a plain listing, full path for a recursive one.

```
$ plakar ls SNAP:/d
0001-01-01T00:00:00Z -rw-r--r--  0  0  5 B readable.txt
0001-01-01T00:00:00Z drwx------  0  0  0 B sub
denied.txt: permission denied

$ plakar ls -recursive SNAP:/d
0001-01-01T00:00:00Z -rw-r--r--  0  0  5 B /d/readable.txt
0001-01-01T00:00:00Z drwx------  0  0  0 B /d/sub
/d/denied.txt: permission denied
/d/sub/nested.txt: permission denied
```

(`denied.txt` is absent from stdout in both: the file failed during the backup, so it never made it into the vfs.)

The issue also says the `errors` subcommand should go away once this is done. I left that out: it removes user-facing surface that now lives under `plakar info -errors`, and it seemed better to have your call on it than to bundle it here. Happy to follow up with it.

Fixes #1268

### DIFF
--- a/subcommands/ls/ls.go
+++ b/subcommands/ls/ls.go
@@ -21,6 +21,8 @@ import (
 	"flag"
 	"fmt"
 	"io/fs"
+	"iter"
+	gopath "path"
 	"strings"
 	"time"
 
@@ -150,8 +152,17 @@ func (cmd *Ls) list_snapshot(ctx *appcontext.AppContext, repo *repository.Reposi
 		return err
 	}
 
+	// Errors recorded during the backup are kept aside from the entries
+	// themselves; list them on stderr, interleaved with the listing.
+	var errs *errorLister
+	defer func() {
+		if errs != nil {
+			errs.close()
+		}
+	}()
+
 	resolved := false
-	return pvfs.WalkDir(pathname, func(path string, d *vfs.Entry, err error) error {
+	err = pvfs.WalkDir(pathname, func(path string, d *vfs.Entry, err error) error {
 		if err != nil {
 			return err
 		}
@@ -165,9 +176,14 @@ func (cmd *Ls) list_snapshot(ctx *appcontext.AppContext, repo *repository.Reposi
 			// right physical path and do our logic on it.
 			resolved = true
 			pathname = d.Path()
+			errs = newErrorLister(pvfs, pathname)
 		}
 		if d.IsDir() && path == pathname {
 			return nil
+		}
+
+		if err := errs.flush(ctx, recursive, pathname, path); err != nil {
+			return err
 		}
 
 		sb, err := d.Info()
@@ -212,4 +228,76 @@ func (cmd *Ls) list_snapshot(ctx *appcontext.AppContext, repo *repository.Reposi
 		}
 		return nil
 	})
+	if err != nil {
+		return err
+	}
+
+	if errs == nil {
+		// the walk yielded nothing, so the lister was never set up
+		errs = newErrorLister(pvfs, pathname)
+	}
+	// entries are exhausted, whatever is left only sorts after them
+	return errs.flush(ctx, recursive, pathname, "")
+}
+
+// errorLister walks the errors recorded in a snapshot, which are sorted by
+// pathname, so that they can be merged into the listing of a directory.
+type errorLister struct {
+	next func() (*vfs.ErrorItem, error, bool)
+	stop func()
+
+	item *vfs.ErrorItem
+	err  error
+}
+
+func newErrorLister(pvfs *vfs.Filesystem, beneath string) *errorLister {
+	next, stop := iter.Pull2(pvfs.Errors(beneath))
+	el := &errorLister{next: next, stop: stop}
+	el.advance()
+	return el
+}
+
+func (el *errorLister) close() {
+	el.stop()
+}
+
+func (el *errorLister) advance() {
+	item, err, ok := el.next()
+	switch {
+	case !ok:
+		el.item = nil
+	case err != nil:
+		el.item, el.err = nil, err
+	default:
+		el.item = item
+	}
+}
+
+// flush reports every error that sorts before upto, or all the remaining ones
+// when upto is empty.  Unless the listing is recursive, errors below beneath
+// are skipped: they belong to entries that are not listed either.
+func (el *errorLister) flush(ctx *appcontext.AppContext, recursive bool, beneath, upto string) error {
+	for el.err == nil && el.item != nil {
+		if upto != "" && el.item.Name >= upto {
+			return nil
+		}
+
+		name := el.item.Name
+		if !recursive {
+			if gopath.Dir(name) != beneath {
+				el.advance()
+				continue
+			}
+			name = gopath.Base(name)
+		}
+
+		fmt.Fprintf(ctx.Stderr, "%s: %s\n", utils.SanitizeText(name),
+			utils.SanitizeText(el.item.Error))
+		el.advance()
+	}
+
+	if el.err != nil {
+		return fmt.Errorf("ls: failed to scan errors: %w", el.err)
+	}
+	return nil
 }
--- a/subcommands/ls/ls_test.go
+++ b/subcommands/ls/ls_test.go
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/PlakarKorp/kloset/connectors"
+	"github.com/PlakarKorp/kloset/objects"
 	"github.com/PlakarKorp/kloset/repository"
 	"github.com/PlakarKorp/kloset/snapshot"
 	"github.com/PlakarKorp/plakar/appcontext"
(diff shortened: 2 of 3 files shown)


poolpOrg:
Ouuuuuh yeah me likey

Will review shortly but I've been wanting to do that for a little while, thanks

mathieu-plak:
Me not likey, and that's why this ticket has been lingering, our team is split on this issue :P. 

I think I could swallow the pill if it was gated behind a flag

omar-polo:
> Me not likey, and that's why this ticket has been lingering, our team is split on this issue :P.

exactly, I'm quite conflicted about this as well.

I'm not sure about using two streams: so far, standard error is for plakar errors, so we should be careful when printing other things to it.

(then personally I think it's the wrong scope, since there's already a command dedicated to the errors in the snapshot)

Anyway, I think I could get used to see them in ls, but please keep using standard output since these are not plakar errors, it's merely printing information from kloset, and if we can avoid adding flags I'd be happier :>

poolpOrg:
If you have issues on a file or directory with 'ls', where is it printed ?

omar-polo:
nothing, because if you can read a directory you can get all the dirents, it's just accessing the file that fails.  the only two situations where ls can fail AFAIK is 1) when the target does not exists or 2) when you pass `-R` and it fails to enter a sub-*directory*.

in our case, 1/ is already handled, and it goes to standard error.  for 2/ we're not failing in any way.  all the data is there, we're just doing a print / replay, so polluting standard error for something that is 100% reproducible feels wrong.

poolpOrg:
That's not what I remember from backing up /private/etc, will test again when on front of a computer but it feels very absurd to me that we need to run an info or diag to list errors that happened during an import FS walk when our 'plakar ls' is supposed to mimic an ls on a frozen snapshot. I feel like I should see what I have seen during backup, success and errors.

Also not showing in ls means you don't realize there's an error if you don't know it's a separate command.

mathieu-plak:
Just gate in on a -E flag and be done with it...

omar-polo:
> That's not what I remember from backing up /private/etc

just to be clear, I was referring to ls(1) output, not plakar ls output with my "you can see the dirents".  (and actually was talking with mathieu that we could / should mimick the same behaviour, since for unreadable files we get a failure at Open() time, but we have the metadata about it at least.)

> Also not showing in ls means you don't realize there's an error if you don't know it's a separate command.

yeah, that's something I can buy.  it's mostly the stderr choice that's making me a bit uncomfortable still, since at this point I'd display them as regular entries (but marked as such.)

> Just gate in on a -E flag and be done with it...

I'd be happier if we don't.  Throwing flags at use-cases just to cover them is a bad design IMHO.  Either we decide errors are important and we always show them, or we disagree and we have `plakar info -errors` for them.  Adding flags to optionally display something feels very backward to me.

mathieu-plak:
> > That's not what I remember from backing up /private/etc
> 
> just to be clear, I was referring to ls(1) output, not plakar ls output with my "you can see the dirents". (and actually was talking with mathieu that we could / should mimick the same behaviour, since for unreadable files we get a failure at Open() time, but we have the metadata about it at least.)
> 
> > Also not showing in ls means you don't realize there's an error if you don't know it's a separate command.
> 
> yeah, that's something I can buy. it's mostly the stderr choice that's making me a bit uncomfortable still, since at this point I'd display them as regular entries (but marked as such.)
> 
> > Just gate in on a -E flag and be done with it...
> 
> I'd be happier if we don't. Throwing flags at use-cases just to cover them is a bad design IMHO. Either we decide errors are important and we always show them, or we disagree and we have `plakar info -errors` for them. Adding flags to optionally display something feels very backward to me.

ls -a ?

0xfandom:
Ran the `/private/etc` case on macOS (arm64, unprivileged), against `7208e7d4` and against this branch.

```
$ plakar at /tmp/etcstore backup /private/etc
eec9cebc: backup completed with 11 errors in 69.605ms (in: 0 B, out: 6.0 MiB)
```

Five of those are directly under `/private/etc`, six below it.

**Before (`7208e7d4`)** — 72 entries on stdout, nothing on stderr:

```
$ plakar at /tmp/etcstore ls eec9cebc:/private/etc | sed -n '28,32p'
2023-12-05T07:37:15Z -r--r--r--     root    wheel  2.4 KiB man.conf
2023-12-05T07:37:15Z -rw-r--r--     root    wheel     36 B manpaths
2023-12-05T07:37:15Z drwxr-xr-x     root    wheel     64 B manpaths.d
2023-12-05T07:37:15Z -rw-r--r--     root    wheel     53 B networks
2023-12-05T07:37:15Z -rw-r--r--     root    wheel  1.3 KiB newsyslog.conf
```

`master.passwd` sorts between `manpaths.d` and `networks`. It isn't there and nothing says why, same for `aliases.db`, `krb5.keytab`, `sudo_lecture` and `sudoers`. You have to already suspect something is wrong to go run `info -errors`.

**After** — the same 72 lines on stdout, byte for byte, plus on stderr:

```
aliases.db: open /private/etc/aliases.db: permission denied
krb5.keytab: open /private/etc/krb5.keytab: permission denied
master.passwd: open /private/etc/master.passwd: permission denied
sudo_lecture: open /private/etc/sudo_lecture: permission denied
sudoers: open /private/etc/sudoers: permission denied
```

@omar-polo, on the two cases where a real `ls` fails: the second one does happen here. `/private/etc/cups/certs` is a directory the walk could not enter, and on `7208e7d4` a recursive listing shows it with no children and no complaint:

```
$ plakar at /tmp/etcstore ls -recursive eec9cebc:/private/etc | grep cups/certs
2026-07-31T09:02:20Z dr-x--x--x      _lp    admin     64 B /private/etc/cups/certs
$ # stderr is empty
```

With the patch, `-recursive` reports all 11 on stderr, `/private/etc/cups/certs: permission denied` among them, and stdout is unchanged (255 lines either way).

To be fair to your point though, that's 1 of the 11. The other 10 are files the importer could not open, which a real `ls` would indeed never have tried to read, so it would have no reason to complain about them.

That last part is what I'd like to keep whatever we decide on the rest: the errors stay on stderr, so the listing on stdout is exactly what it was before and nothing downstream of a pipe notices.

poolpOrg:
> > Just gate in on a -E flag and be done with it...
> 
> 
> 
> I'd be happier if we don't.  Throwing flags at use-cases just to cover them is a bad design IMHO.  Either we decide errors are important and we always show them, or we disagree and we have `plakar info -errors` for them.  Adding flags to optionally display something feels very backward to me.

Not showing the errors unless the user remembers to check if there are any (with a flag or even worse with a different subcommand) is a guaranteed way to have users realize their snapshot is incomplete the day they try to restore it.

Do you guys genuinely think that people run both ´plakar ls' and plakar info -errors' after creating a snapshot to check if it has content and errors ?

I don't understand the issue with outputting errors on stderr so they are visible.

mathieu-plak:
> > > Just gate in on a -E flag and be done with it...
> > 
> > 
> > I'd be happier if we don't.  Throwing flags at use-cases just to cover them is a bad design IMHO.  Either we decide errors are important and we always show them, or we disagree and we have `plakar info -errors` for them.  Adding flags to optionally display something feels very backward to me.
> 
> Not showing the errors unless the user remembers to check if there are any (with a flag or even worse with a different subcommand) is a guaranteed way to have users realize their snapshot is incomplete the day they try to restore it.
> 
> Do you guys genuinely think that people run both ´plakar ls' and plakar info -errors' after creating a snapshot to check if it has content and errors ?
> 
> I don't understand the issue with outputting errors on stderr so they are visible.

The way we (I assume op@ shares the same sentiment) see it is that errors are data. We expect ls to output stuff on stderr for errors happening during the ls (corrupted packfile etc).

As for users well isn't backup time the moment you catch up the error? 

Just giving my reasoning, this is not a hill I will die on.

poolpOrg:
> > Not showing the errors unless the user remembers to check if there are any (with a flag or even worse with a different subcommand) is a guaranteed way to have users realize their snapshot is incomplete the day they try to restore it.
> > Do you guys genuinely think that people run both ´plakar ls' and plakar info -errors' after creating a snapshot to check if it has content and errors ?
> > I don't understand the issue with outputting errors on stderr so they are visible.
> 
> The way we (I assume op@ shares the same sentiment) see it is that errors are data. We expect ls to output stuff on stderr for errors happening during the ls (corrupted packfile etc).
> 

I understand your point, but I think it's highly biased toward a developer-centric view of the semantics rather than what is most intuitive and helpful for users.

From a user's perspective, an error encountered during a backup isn't "just data" because we have decided to record them in a b-tree for persistence, it's still an error at the snapshot level. The fact that we choose to display it later rather than immediately doesn't change its nature. Because of that, it makes sense for these messages to go to `stderr`, even if they're being replayed as part of a deferred display.

This has already confused people several times, starting with our own team (I think Niluje first questioned it, then me), and later with lambda users who were surprised to discover `plakar info -errors` existed when they asked why `plakar ls` was not showing the errors.

I can predict what will happen if we leave things as they are: eventually someone will restore a snapshot, notice that files are missing, and wonder why `plakar ls` gave no indication that anything went wrong. We'll then have to explain that for us, developers, the recorded backup errors aren't really "errors" anymore, they've become "data", because we've reserved the notion of errors for packfile-related issues and not previous errors that we recorded. I don't think that's a distinction lambda users will naturally make, understand, or be accept as a good rationale.

Personally, I think of `plakar ls` as replaying what `ls` would have shown at the moment the snapshot was taken (that is the FS walker). If errors occurred during that listing and we recorded them, it feels inconsistent not to surface them when replaying it.


> As for users well isn't backup time the moment you catch up the error?
> 

Yes but still, if they somehow missed it, why punish them ?

If we have a way to hint that there's an issue with a snapshot content,
we should always try to make this surface so they will see it as soon as possible,
not when they need to restore it and realize they are fuckd.

We are in data protection, not data protection education :-)


> Just giving my reasoning, this is not a hill I will die on.

Good cause I would haunt your dreams.

mathieu-plak:
> > > Not showing the errors unless the user remembers to check if there are any (with a flag or even worse with a different subcommand) is a guaranteed way to have users realize their snapshot is incomplete the day they try to restore it.
> > > Do you guys genuinely think that people run both ´plakar ls' and plakar info -errors' after creating a snapshot to check if it has content and errors ?
> > > I don't understand the issue with outputting errors on stderr so they are visible.
> > 
> > 
> > The way we (I assume op@ shares the same sentiment) see it is that errors are data. We expect ls to output stuff on stderr for errors happening during the ls (corrupted packfile etc).
> 
> I understand your point, but I think it's highly biased toward a developer-centric view of the semantics rather than what is most intuitive and helpful for users.
> 
> From a user's perspective, an error encountered during a backup isn't "just data" because we have decided to record them in a b-tree for persistence, it's still an error at the snapshot level. The fact that we choose to display it later rather than immediately doesn't change its nature. Because of that, it makes sense for these messages to go to `stderr`, even if they're being replayed as part of a deferred display.
> 
> This has already confused people several times, starting with our own team (I think Niluje first questioned it, then me), and later with lambda users who were surprised to discover `plakar info -errors` existed when they asked why `plakar ls` was not showing the errors.
> 
> I can predict what will happen if we leave things as they are: eventually someone will restore a snapshot, notice that files are missing, and wonder why `plakar ls` gave no indication that anything went wrong. We'll then have to explain that for us, developers, the recorded backup errors aren't really "errors" anymore, they've become "data", because we've reserved the notion of errors for packfile-related issues and not previous errors that we recorded. I don't think that's a distinction lambda users will naturally make, understand, or be accept as a good rationale.
> 
> Personally, I think of `plakar ls` as replaying what `ls` would have shown at the moment the snapshot was taken (that is the FS walker). If errors occurred during that listing and we recorded them, it feels inconsistent not to surface them when replaying it.
> 
> > As for users well isn't backup time the moment you catch up the error?
> 
> Yes but still, if they somehow missed it, why punish them ?
> 
> If we have a way to hint that there's an issue with a snapshot content, we should always try to make this surface so they will see it as soon as possible, not when they need to restore it and realize they are fuckd.
> 
> We are in data protection, not data protection education :-)
> 
> > Just giving my reasoning, this is not a hill I will die on.
> 
> Good cause I would haunt your dreams.

You got me "convinced" (aka I was outranked :>). No but for real okay let's try to get this merged. I need to do a proper review, from a cursory reading this was a tad more complex than I would have imagined.

poolpOrg (approved):

PR #2297, opened 2026-07-31
api: keep partial vfs search with mime filters

### LINKED ISSUE #1304: api: vfssearch does full match with mime but not without
searching for `CODE` yield stuff like `CODE_OF_CONDUCT` *but* if there is a mime filter, then the check becomes strict and one has to use `CODE*` instead.

ousamabenyounes:
## Summary
- preserve partial filename matching for `/api/snapshot/vfs/search` when a MIME filter is also present
- bypass the MIME-index search path only for `pattern + mime`, then apply MIME filtering in the API before pagination
- add a regression fixture and test for `CODE` matching `CODE_OF_CONDUCT` with `mime=text/plain`

Fix #1304

RED on upstream behavior with the new API regression test:

```text
--- FAIL: TestSnapshotVFSSearchMimeFilterKeepsPartialNameMatch (0.19s)
    api_snapshot_test.go:117:
        	Error:      	"{\"has_next\":false,\"items\":[]}\n" does not contain "CODE_OF_CONDUCT"
FAIL
FAIL	github.com/PlakarKorp/plakar/api	0.208s
```

GREEN targeted after the fix:

```text
ok  	github.com/PlakarKorp/plakar/api	0.200s
```

Package and full local CI replay:

```text
go test ./api -count=1
ok  	github.com/PlakarKorp/plakar/api	13.599s

origin/main baseline at 693723b4: build, vet, and full tests PASS
patched branch at a9606108:      build, vet, and full tests PASS
local replay note: both trees report the same pre-existing gofmt file list
patched focused regression:      ok github.com/PlakarKorp/plakar/api 0.157s
```

### DIFF
--- a/api/helpers_test.go
+++ b/api/helpers_test.go
@@ -17,6 +17,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const searchFixtureConductFile = "CODE_OF_CONDUCT"
+
 // server builds a real repository with a richer snapshot than newAPIServer
 // (nested dirs, multiple files) and wires SetupRoutes with norefresh=true.
 // Distinct name to avoid clashing with the team's newAPIServer helper.
@@ -31,6 +33,7 @@ func server(t *testing.T, token string) (*http.ServeMux, *repository.Repository,
 		ptesting.NewMockDir("subdir/nested"),
 		ptesting.NewMockFile("subdir/nested/deep.txt", 0644, "deep content"),
 		ptesting.NewMockFile("top.txt", 0644, "top level"),
+		ptesting.NewMockFile(searchFixtureConductFile, 0644, "project conduct"),
 	})
 	mux := http.NewServeMux()
 	SetupRoutes(mux, repo, ctx, token, true /* norefresh */)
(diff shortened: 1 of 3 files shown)


api/helpers_test.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const searchFixtureConductFile = "CODE_OF_CONDUCT"
```
omar-polo: I'm sorry, I'm commenting this on almost every PR you've opened, but please can you stop adding constants that have only one use?  It's just making everything harder to read.

omar-polo (changes requested):
Hello!

Sorry for the delay, I was working on something else and then took a week off.  I'm leaving just a cosmetic comment for now, I'll need a bit more time to look into this.

thanks!

api/helpers_test.go
ousamabenyounes: Hi @omar-polo, hope you had a great week off! We’re reviewing all your feedback and working through the fixes now.

ousamabenyounes:
Thanks for the review @omar-polo. Addressed the request in `a9606108` by inlining every one-use constant introduced by this PR; the reused media-type separator remains named.

Validation: the focused API regression test is GREEN, baseline/final `run-ci.sh` results are iso-baseline (build, vet, and full tests pass; both show the same pre-existing formatting list), and all GitHub checks pass.

PR #2298, opened 2026-07-31
api: list repository sources

### LINKED ISSUE #1223: Create an API endpoint listing the distinct sources in a Kloset
The dashboard on the UI currently displays the following graphs:

Image

We want to add a third graph with the following info:

* Title: `Sources`
* Value: `XX distinct sources` (eg. "23 distinct sources") - a source being the unique triplet importer type/origin/directory
* Sub text: `Using XXX importers` (eg. "Using 3 importers", referring to for example fs, sftp, s3)

To do so, we need a new API returning the unique number of sources in the Kloset.

The unique number of importers can already be fetched from the API endpoint `/api/repository/importer-types`.

ousamabenyounes:
## Summary

- Add `GET /api/repository/sources`.
- Return the distinct snapshot source triplets as `{type, origin, directory}` entries.
- Keep the response shape consistent with the other repository list endpoints: `total` plus `items`.
- Endpoint and test values are inlined, as requested in review — the diff introduces no named constant.
- Covers refresh, iterator, load-error, deduplication and ordering paths.

Rebased onto current `main` and squashed into a single commit, so the PR diff is the whole change.

| Q | A |
| --- | --- |
| Issues | Fix #1223 |
| Tests | `go test ./api -run TestRepository -count=1`; full local replay of `go.yml` + `go-windows.yml` |

RED — probe on unmodified `origin/main`, the route does not exist:

```text
=== RUN   TestRepositorySourcesRED
    Error:      	Not equal:
    	            	expected: 200
    	            	actual  : 404
    	Messages:   	body={"error":{"code":"not-found","message":"API endpoint not found"}}
--- FAIL: TestRepositorySourcesRED (0.14s)
FAIL	github.com/PlakarKorp/plakar/api	0.162s
```

GREEN — with this change:

```text
--- PASS: TestRepository/sources (0.26s)
--- PASS: TestRepository/sources_refresh_error (0.07s)
--- PASS: TestRepository/sources_snapshot_load_error (0.18s)
--- PASS: TestRepository/sources_snapshot_list_error (0.14s)
ok  	github.com/PlakarKorp/plakar/api	0.881s
```

Full local CI replay (`go build -v ./...`, `go build tool`, `GOOS=windows` build, `go vet`, `timeout -s QUIT 5m go test -p 4 -coverprofile -covermode=atomic ./...`) on the upstream baseline and on this commit:

```text
baseline (origin/main): 1289 passed, 0 failed
final   (this commit):  1293 passed, 0 failed
diff-coverage-gate: PASS 100.00% (39/39) changed production lines covered
```

## Open point

Left the snapshot cap out on purpose: it changes what the endpoint reports, and the bound depends on what the dashboard actually needs — happy to add it once @brmzkw / @decarheil weigh in on the format.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -197,6 +197,7 @@ func SetupRoutes(server *http.ServeMux, repo *repository.Repository, ctx *appcon
 	server.Handle("GET /api/repository/snapshots", authToken(JSONAPIView(ui.repositorySnapshots)))
 	server.Handle("GET /api/repository/locate-pathname", authToken(JSONAPIView(ui.repositoryLocatePathname)))
 	server.Handle("GET /api/repository/importer-types", authToken(JSONAPIView(ui.repositoryImporterTypes)))
+	server.Handle("GET /api/repository/sources", authToken(JSONAPIView(ui.repositorySources)))
 
 	server.Handle("GET /api/snapshot/{snapshot}", authToken(JSONAPIView(ui.snapshotHeader)))
 	server.Handle("GET /api/snapshot/reader/{snapshot_path...}", urlSigner.VerifyMiddleware(APIView(ui.snapshotReader)))
--- a/api/api_repository_test.go
+++ b/api/api_repository_test.go
@@ -4,8 +4,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"net/http"
+	"path/filepath"
 	"testing"
 
+	"github.com/PlakarKorp/kloset/caching"
+	"github.com/PlakarKorp/kloset/connectors/storage"
+	"github.com/PlakarKorp/kloset/objects"
+	"github.com/PlakarKorp/kloset/resources"
 	ptesting "github.com/PlakarKorp/plakar/testing"
 	"github.com/stretchr/testify/require"
 )
@@ -20,6 +25,7 @@ func TestRepositoryEmpty(t *testing.T) {
 		for _, path := range []string{
 			"/api/repository/snapshots",
 			"/api/repository/importer-types",
+			"/api/repository/sources",
 			"/api/repository/locate-pathname",
 			"/api/repository/info",
 		} {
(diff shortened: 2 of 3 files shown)


api/api.go
```diff
@@ -199,6 +201,7 @@ func SetupRoutes(server *http.ServeMux, repo *repository.Repository, ctx *appcon
 	server.Handle("GET /api/repository/snapshots", authToken(JSONAPIView(ui.repositorySnapshots)))
 	server.Handle("GET /api/repository/locate-pathname", authToken(JSONAPIView(ui.repositoryLocatePathname)))
 	server.Handle("GET /api/repository/importer-types", authToken(JSONAPIView(ui.repositoryImporterTypes)))
+	server.Handle("GET "+repositorySourcesPath, authToken(JSONAPIView(ui.repositorySources)))
```
omar-polo: please don't, it's fine to hardcode the endpoints

api/api_repository_test.go
```diff
@@ -22,6 +24,7 @@ func TestRepositoryEmptyStateOK(t *testing.T) {
 	for _, path := range []string{
 		"/api/repository/snapshots",
 		"/api/repository/importer-types",
+		repositorySourcesPath,
```
omar-polo: ditto, introducing constants for just one or two uses it's not something we're doing.  please inline.

omar-polo (changes requested):
hello =)

leaving some nits here.  the only thing i'm not sure is that the `type+origin+root` can make for quite some entries due to origin and root explosions, so maybe we want to put a cap on the maximum number of snapshot we're going to consider?

leaving this more as an open question than anything.  it'll also need some input from @brmzkw or @decarheil to see if this format could actually be useful in the UI.

ousamabenyounes:
Thanks @omar-polo. Addressed both in 20c29b6a: dropped the `repositorySourcesPath` constant and inlined the endpoint (matching the surrounding `server.Handle` calls), and inlined the test constants (`repositorySourcesPath` uses + the one-off `repositorySourcesSecondSnapshotName`). `go build ./api/...` and `go test ./api` stay green.

On the entry-count concern: happy to add a cap on the sources listing if you'd like — let me know the bound you'd prefer and I'll fold it in.

poolpOrg:
Why Directory and not Root ?

ousamabenyounes:
I named it `directory` because this endpoint is a direct projection of the persisted importer metadata in the snapshot header: `Importer.Type`, `Importer.Origin`, `Importer.Directory`.

`root` is already used for other concepts here, notably the VFS/repository root, so reusing it for the importer source path would be ambiguous. Also, the existing `locate-pathname` endpoint already exposes the same importer triple through `importerType`, `importerOrigin`, and `importerDirectory`, so `directory` keeps this endpoint consistent with the current API.

If you prefer `root` for UI terminology, I am happy to rename it before merge, but I would lean toward keeping it aligned with the underlying importer field and existing query params.

omar-polo (changes requested):
you've marked my comments as fixed but the constant is still there ^^;

ousamabenyounes:
Thank you for the reviews. The one-use endpoint and test constants are now inlined, and the branch has been rebased onto the current `main` in e6048818ca30b5062b96b190c0fae26a3084c557.

Validation: the endpoint test reproduced the missing route on the upstream baseline (RED), then the focused test and full local CI baseline/final replay passed (GREEN); all visible GitHub checks are also successful.

ousamabenyounes:
@omar-polo sorry for the confusion — I think the "constant is still there" you saw was the *outdated* hunk still rendered inside the two resolved conversations; the branch itself had already dropped it. To make that unambiguous I rebased onto current `main` and squashed the three commits into a single `api: list repository sources` (655a4b6dfb385f4ab5cf950cd6b1a100737b47a8), so the PR diff now *is* the whole change. `grep -n const api/api.go api/api_repository.go api/api_repository_test.go` comes back empty; the route is inline like its neighbours:

```go
server.Handle("GET /api/repository/sources", authToken(JSONAPIView(ui.repositorySources)))
```

On the cardinality question: I left it out on purpose. A cap changes what the endpoint means (the dashboard counter in #1223 would silently under-report), and the bound only makes sense once @brmzkw / @decarheil confirm the shape they want in the UI. Happy to add it in a follow-up — or here — once there's a number to pin.

@poolpOrg on `directory` vs `root`: still happy to rename before merge if you prefer the UI wording. I kept `directory` because it mirrors the persisted `Importer.Directory` field and the existing `locate-pathname` query params (`importerType` / `importerOrigin` / `importerDirectory`), and `root` is already taken by the VFS/repository root here.

Validation: RED on unmodified `origin/main` (`expected: 200; actual: 404`, `body={"error":{"code":"not-found",...}}`), GREEN after the change. Full local CI replay of `go.yml` + `go-windows.yml` on the upstream baseline and on this commit: 1289 → 1293 tests passing, 0 failing on both, no new failure signature.

PR #2299, opened 2026-07-31
login: report auth rate limits as 429

### LINKED ISSUE #1656: plakar login: unexpected 500 error after shell interruption
Hi team,

I encountered a bug when trying to authenticate using the plakar login command. Here are the steps to reproduce it:

Steps to Reproduce

- Run plakar login from the terminal
- Interrupt the login process with CTRL+C while waiting for browser-based confirmation
- Click on the email button in the terminal to re-initiate the login
- Try logging in again

Observed Behavior

The command fails with the following error:

plakar: unexpected status code: 500 : {"error":"OpError[limit-reached] Limit reached"}

Expected Behavior

The CLI should allow the login attempt after an aborted flow. Ideally, the authentication session should be reset automatically if it was interrupted by the user.

Environment

OS: Ubuntu

Plakar version: v1.0.3.rc2

Additional Notes

The issue seems to be related to rate limiting or session locking. Happy to help test a patch or provide more logs if needed.

Thanks for the great work on Plakar, loving the tool so far! :)

ousamabenyounes:
Fix #1656

## Summary
When the auth API rate-limits a login attempt it replies `500` with a body
containing `limit-reached` (see the issue). The CLI/UI previously surfaced this
as an opaque `unexpected status code: 500 : {...}` error. This maps that
condition to a `429` `rate-limited` API response.

Per review, the mechanism is a package-level sentinel rather than an exported
HTTP-shaped error type:

- `login.ErrRateLimited` is wrapped into the returned error when a login
  response is rate-limited (429 status, or a body containing `limit-reached`).
- The API layer detects it with `errors.Is(err, login.ErrRateLimited)` and
  returns a `429` `rate-limited` `ApiError`.
- Response bodies are now closed on every `Poll` branch, and the `Run`/`RunUI`
  `defer resp.Body.Close()` calls are back in their original position;
  `unexpectedStatusError` only reads the body and leaves closing to the caller.

The earlier `NewLoginFlow` `PLAKAR_API_URL` change has been reverted — it only
existed to point an integration test at a local server. The behaviour is now
covered by focused unit tests that need no such override.

RED — with the rate-limit detection/mapping removed, the new tests fail,
reproducing the issue's exact payload:

```text
--- FAIL: TestRunWrapsErrRateLimitedFromBodyMarker
    login_ratelimit_test.go:26: err = unexpected status code: 500 : {"error":"OpError[limit-reached] Limit reached"}, want wrapping ErrRateLimited
--- FAIL: TestRunWrapsErrRateLimitedFromStatusCode
    login_ratelimit_test.go:41: err = unexpected status code: 429 : , want wrapping ErrRateLimited
--- FAIL: TestLoginFlowErrorMapsRateLimitTo429
    api_test.go:130: Should be in error chain: expected *api.ApiError
```

GREEN — with the fix:

```text
ok  github.com/PlakarKorp/plakar/login
ok  github.com/PlakarKorp/plakar/api
```

Full local CI (`go build -v ./...` and
`go test -p 4 -coverprofile=coverage.txt -covermode=atomic ./...`) is
iso-baseline with `main`.

### DIFF
--- a/api/api_authentication.go
+++ b/api/api_authentication.go
@@ -40,7 +40,7 @@ func (ui *uiserver) servicesLoginGithub(w http.ResponseWriter, r *http.Request) 
 
 	redirectURL, err := lf.RunUI("github", parameters)
 	if err != nil {
-		return fmt.Errorf("failed to run login flow: %w", err)
+		return loginFlowError(err)
 	}
 
 	ret := struct {
@@ -70,7 +70,7 @@ func (ui *uiserver) servicesLoginEmail(w http.ResponseWriter, r *http.Request) e
 
 	redirectURL, err := lf.RunUI("email", parameters)
 	if err != nil {
-		return fmt.Errorf("failed to run login flow: %w", err)
+		return loginFlowError(err)
 	}
 
 	ret := struct {
@@ -79,6 +79,18 @@ func (ui *uiserver) servicesLoginEmail(w http.ResponseWriter, r *http.Request) e
 		URL: redirectURL,
 	}
 	return json.NewEncoder(w).Encode(ret)
+}
+
+func loginFlowError(err error) error {
+	var statusErr *login.HTTPStatusError
+	if errors.As(err, &statusErr) && statusErr.RateLimited() {
+		return &ApiError{
+			HttpCode: http.StatusTooManyRequests,
+			ErrCode:  "rate-limited",
+			Message:  "Rate limit reached. Please retry later.",
+		}
+	}
+	return fmt.Errorf("failed to run login flow: %w", err)
 }
 
 func (ui *uiserver) servicesLogout(w http.ResponseWriter, r *http.Request) error {
--- a/api/api_authentication.go
+++ b/api/api_authentication.go
@@ -82,8 +82,7 @@ func (ui *uiserver) servicesLoginEmail(w http.ResponseWriter, r *http.Request) e
 }
 
 func loginFlowError(err error) error {
-	var statusErr *login.HTTPStatusError
-	if errors.As(err, &statusErr) && statusErr.RateLimited() {
+	if errors.Is(err, login.ErrRateLimited) {
 		return &ApiError{
 			HttpCode: http.StatusTooManyRequests,
 			ErrCode:  "rate-limited",
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -4,20 +4,19 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io/fs"
 	"net/http"
 	"net/http/httptest"
-	"strings"
 	"testing"
 
 	_ "github.com/PlakarKorp/integrations/fs/exporter"
 	"github.com/PlakarKorp/kloset/repository"
 	"github.com/PlakarKorp/kloset/snapshot"
+	"github.com/PlakarKorp/plakar/login"
 	ptesting "github.com/PlakarKorp/plakar/testing"
 	"github.com/stretchr/testify/require"
 )
-
-const loginLimitReachedResponse = `{"error":"OpError[limit-reached] Limit reached"}`
 
 func TestHandleErrorMapping(t *testing.T) {
 	cases := []struct {
@@ -124,25 +123,19 @@ func TestApiInfoAuthenticated(t *testing.T) {
 	require.NotEmpty(t, resp.RepositoryId)
 }
 
-func TestServicesLoginEmailRateLimitReturnsTooManyRequests(t *testing.T) {
-	authServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		require.Equal(t, "/v1/auth/login/email", r.URL.Path)
-		w.WriteHeader(http.StatusInternalServerError)
-		_, err := w.Write([]byte(loginLimitReachedResponse))
-		require.NoError(t, err)
-	}))
-	defer authServer.Close()
+func TestLoginFlowErrorMapsRateLimitTo429(t *testing.T) {
+	err := loginFlowError(fmt.Errorf("failed to run login flow: %w", login.ErrRateLimited))
 
-	t.Setenv("PLAKAR_API_URL", authServer.URL)
+	var apiErr *ApiError
+	require.ErrorAs(t, err, &apiErr)
+	require.Equal(t, http.StatusTooManyRequests, apiErr.HttpCode)
+	require.Equal(t, "rate-limited", apiErr.ErrCode)
+}
 
-	mux, _, snap, _ := server(t, "")
-	defer snap.Close()
+func TestLoginFlowErrorPassesThroughOtherErrors(t *testing.T) {
+	err := loginFlowError(errors.New("boom"))
 
-	req, err := http.NewRequest("POST", "/api/authentication/login/email", strings.NewReader(`{"email":"user@example.com"}`))
-	require.NoError(t, err)
-	w := httptest.NewRecorder()
-	mux.ServeHTTP(w, req)
-
-	require.Equal(t, http.StatusTooManyRequests, w.Code, "body=%s", w.Body.String())
-	require.Contains(t, w.Body.String(), "rate-limited")
+	var apiErr *ApiError
+	require.False(t, errors.As(err, &apiErr), "non-rate-limit error must not be mapped to an ApiError")
+	require.ErrorContains(t, err, "boom")
 }
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -7,6 +7,7 @@ import (
 	"io/fs"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 
 	_ "github.com/PlakarKorp/integrations/fs/exporter"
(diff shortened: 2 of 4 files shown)


login/login.go
```diff
@@ -38,21 +39,55 @@ type TokenResponse struct {
...
+	Body       string
+}
+
+func (err *HTTPStatusError) Error() string {
+	return fmt.Sprintf("unexpected status code: %d : %s", err.StatusCode, err.Body)
+}
+
+func (err *HTTPStatusError) RateLimited() bool {
+	return err.StatusCode == http.StatusTooManyRequests || strings.Contains(err.Body, rateLimitErrorMarker)
+}
```
omar-polo: I think this is the wrong abstraction for the layer at hand.  Instead of doing this, I'd just introduce a global error in this package, say `ErrRateLimited` and check against that in the calling code.  e.g.

```go
package login

var ErrRateLitimed = errors.New("rate limited")

func Poll(...) {
  // [...]
  if err != nil {
    return fmt.Errorf("login failed: %w: %w", ErrRateLimited, err)
  }
}

// ====

package api

func handler(..) {
  if err := login.Poll(); err != nil {
    if errors.Is(err, login.ErrRateLimited) {
      return 429
    }
    return 500
  }
}
```

login/login.go
```diff
@@ -38,21 +39,55 @@ type TokenResponse struct {
...
 	appCtx  *appcontext.AppContext
 	noSpawn bool
 	baseURL string
 }
 
 func NewLoginFlow(appCtx *appcontext.AppContext, noSpawn bool) (*loginFlow, error) {
+	baseURL := defaultBaseURL
+	if envURL := os.Getenv(plakarAPIURLEnv); envURL != "" {
+		baseURL = envURL
+	}
```
omar-polo: this is a change in behaviour, why?

login/login.go
```diff
@@ -71,20 +106,22 @@ func (flow *loginFlow) Poll(pollID string, iterations int, delay time.Duration,
 			if err != nil {
 				return "", fmt.Errorf("the /auth/login/github/poll API endpoint failed: %w", err)
 			}
-			// leaking resp for now
```
omar-polo, line 74: but it's still leaked, no?

login/login.go
```diff
@@ -115,12 +152,11 @@ func (flow *loginFlow) Run(provider string, parameters map[string]string) (strin
...
 		return "", fmt.Errorf("unable to get the login URL: %w", err)
 	}
-	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		data, _ := io.ReadAll(resp.Body)
-		return "", fmt.Errorf("unexpected status code: %d : %s", resp.StatusCode, data)
+		return "", unexpectedStatusError(resp)
 	}
+	defer resp.Body.Close()
```
omar-polo: why moving the defer `Body.Close()`?

login/login.go
```diff
@@ -192,12 +228,11 @@ func (flow *loginFlow) RunUI(provider string, parameters map[string]string) (str
...
```
omar-polo: same question here as well

omar-polo (commented):
hello.

thanks!  overall it's good, there is just a remark i'm leaving behind and some questions.  I think that exposing an `HTTPError` in the login package is the wrong abstraction: a dedicated error seems more in line with the usual way known failure modes are reported in go.  Then, some minor nits regarding the closing of bodies.

ousamabenyounes:
Thanks for the review @omar-polo. Reworked per your comments in da4cd8587f9c:

- Dropped the exported `HTTPStatusError`/`RateLimited()` in favour of a
  package-level `login.ErrRateLimited` sentinel; the api handler maps it via
  `errors.Is(err, login.ErrRateLimited)` -> 429, as you suggested.
- Reverted the `NewLoginFlow` `PLAKAR_API_URL` behaviour change — it only
  existed to point an integration test at a local server. The mapping is now
  covered by focused unit tests in the `login` and `api` packages, so the
  production behaviour change is gone.
- `Poll` now closes `resp.Body` on every branch (no defer-in-loop), and the
  `Run`/`RunUI` `defer resp.Body.Close()` are back where they were;
  `unexpectedStatusError` only reads the body and leaves closing to the caller.

Validation: reproduced #1656's payload (`unexpected status code: 500 :
{...limit-reached...}`) as a failing test, then green after the fix; full local
suite iso-baseline with `main`.

login/login.go
ousamabenyounes: Done — replaced the exported `HTTPStatusError`/`RateLimited()` with a package-level `login.ErrRateLimited` sentinel, checked with `errors.Is` in the api layer exactly as you sketched. No HTTP-shaped error type is exported anymore.

login/login.go
ousamabenyounes: Reverted — `NewLoginFlow` no longer reads `PLAKAR_API_URL`. That was only there to route the integration test at a local server; the rate-limit mapping is now unit-tested in the `login` and `api` packages without any production behaviour change.

login/login.go
ousamabenyounes: Fixed — `Poll` now closes `resp.Body` on all four branches (OK / NotFound / Accepted / default), with no `defer`-in-loop. The default branch closes the body after `unexpectedStatusError` has read it.

login/login.go
ousamabenyounes: Reverted to the original position (right after the POST). `unexpectedStatusError` no longer closes the body, so the single `defer` covers both the OK and error paths with no double-close.

login/login.go
ousamabenyounes: Same as Run — `defer resp.Body.Close()` is back in its original position and `unexpectedStatusError` only reads the body.

mathieu-plak (approved):

poolpOrg (approved):

api/api_authentication.go
```diff
@@ -40,7 +40,7 @@ func (ui *uiserver) servicesLoginGithub(w http.ResponseWriter, r *http.Request)
 
 	redirectURL, err := lf.RunUI("github", parameters)
 	if err != nil {
-		return fmt.Errorf("failed to run login flow: %w", err)
+		return loginFlowError(err)
```
omar-polo, line 43: you can keep the existing `fmt.Errorf` here -- is correct.

the right place to check for errors is not yet another function but rather `handleError` in api.go: just add an arm there to catch and map login flow errors.

api/api_authentication.go
```diff
@@ -70,7 +70,7 @@ func (ui *uiserver) servicesLoginEmail(w http.ResponseWriter, r *http.Request) e
 
 	redirectURL, err := lf.RunUI("email", parameters)
 	if err != nil {
-		return fmt.Errorf("failed to run login flow: %w", err)
+		return loginFlowError(err)
```
omar-polo, line 73: ditto

api/api_test.go
```diff
@@ -120,3 +122,20 @@ func TestApiInfoAuthenticated(t *testing.T) {
...
+	require.Equal(t, "rate-limited", apiErr.ErrCode)
+}
+
+func TestLoginFlowErrorPassesThroughOtherErrors(t *testing.T) {
+	err := loginFlowError(errors.New("boom"))
+
+	var apiErr *ApiError
+	require.False(t, errors.As(err, &apiErr), "non-rate-limit error must not be mapped to an ApiError")
+	require.ErrorContains(t, err, "boom")
+}
```
omar-polo, line 141: ...at which point we don't need the two tests.

login/login.go
```diff
@@ -38,6 +40,13 @@ type TokenResponse struct {
 // (via the baseURL field) so tests can point the login flow at a local server.
 const defaultBaseURL = "https://api.plakar.io"
 
+const rateLimitErrorMarker = "limit-reached"
```
omar-polo, line 43: please, avoid that

login/login.go
```diff
@@ -53,6 +62,22 @@ func NewLoginFlow(appCtx *appcontext.AppContext, noSpawn bool) (*loginFlow, erro
...
+// non-OK response. It does not close resp.Body; the caller owns closing it.
+// When the response signals rate limiting, the returned error wraps
+// ErrRateLimited so callers can detect it via errors.Is.
+func unexpectedStatusError(resp *http.Response) error {
+	data, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("failed to read error response body: %w", err)
+	}
+	statusErr := fmt.Errorf("unexpected status code: %d : %s", resp.StatusCode, data)
+	if resp.StatusCode == http.StatusTooManyRequests || strings.Contains(string(data), rateLimitErrorMarker) {
```
omar-polo, line 75: the status code is enough for this decision, let's no try to match on the content string.

if the server returns the wrong status code, it's a bug on the server and we'll fix it.

login/login.go
```diff
@@ -53,6 +62,22 @@ func NewLoginFlow(appCtx *appcontext.AppContext, noSpawn bool) (*loginFlow, erro
 	return flow, nil
 }
 
+// unexpectedStatusError reads resp.Body to build a descriptive error for a
+// non-OK response. It does not close resp.Body; the caller owns closing it.
+// When the response signals rate limiting, the returned error wraps
+// ErrRateLimited so callers can detect it via errors.Is.
+func unexpectedStatusError(resp *http.Response) error {
+	data, err := io.ReadAll(resp.Body)
```
omar-polo, line 70: also, `io.ReadAll` is a silent trap: a malicious server could send an infinite stream.  let's ingore completely the response body.

(the question of adding some protection to the other case where we're just unmarshaling code is an already existing issue, and let's keep it for another PR eventually)

omar-polo (changes requested):
sorry for the late reply, between my vacations and other things this fell thru the cracks.

omar-polo:
and i haven't realized that this was merged lol.

github ui is amazing.

i'll fix my nits in another PR then, thanks!

PR #2306, opened 2026-08-12
Poolp org/cobra conversion

poolpOrg:


mathieu-plak (approved):

subcommands/cobra_test.go
```diff
@@ -0,0 +1,128 @@
...
+		{
+			name: "one letter long option with equals value",
+			in:   []string{"-o=out.txt"},
+			want: []string{"--o=out.txt"},
+		},
+		{
+			name: "one letter long boolean option",
+			in:   []string{"-u"},
+			want: []string{"--u"},
+		},
```
omar-polo, line 63: this however looks wrong to me: `getopt_long(3)` treats `-o` as a short flag, and `--o` as a run of the option `-` and the option `o`.  Admittedly, I haven't taken a look at what cobra does, but as a user I'd find this very confusing if/when we'll move to cobra only and remove this.

subcommands/subcommands.go
```diff
@@ -41,6 +44,80 @@ func (cmd *SubcommandBase) GetRepositorySecret() []byte {
 	return cmd.RepositorySecret
 }
 
+// SingleDash rewrites our single-dash options into the double-dash form pflag
+// expects, which it would otherwise read as bundles of shorthands.  Length is
+// not the criterion: -o, -k and -u are declared with StringVar and friends, so
+// they are one-letter long options rather than shorthands.
+//
+// Like the flag package it stops at the first positional argument, so a path
+// starting with a dash is never taken for an option.
```
omar-polo: that's a bit misleading.  `plakar backup -foo/bar` will still consider the "path" to be an option, the second part of the sentence does not make sense to me ^^;

(it'd be clearer to either drop it or say what the function does, which is to stop at `--` or the first positional argument)

cmd.go
```diff
@@ -0,0 +1,179 @@
...
+	return f.Value.Type() != "bool"
+}
+
+// normalizeArgs rewrites our command line into one pflag can parse: it spells
+// the single-dash options with two dashes and pulls out the "at REPOSITORY"
+// prefix, which cobra has no notion of.
+//
+// out is the global part, rest is everything from the subcommand name onwards
+// and is never rewritten: the subcommand parses its own options.
+func normalizeArgs(root *cobra.Command, args []string) (out, rest []string, repository string, hadAt bool, err error) {
```
omar-polo, line 65: nit: can we dedup this somehow with the SingleFlag in subcommands?  they share most of the logic and it's tricky anyway for wanting it in a single place IMHO.

for example, can't this be rewritten as a call to SingleFlag, then check for `at ...`, then run SingleFlag again on the rest?

main.go
```diff
@@ -322,8 +327,20 @@ func entryPoint() int {
...
+	// An "at" with nothing behind it is a mistake of its own.
+	if hadAt {
+		if atRepository == "" {
+			fmt.Fprintf(os.Stderr, "%s: missing plakar repository\n", progName())
+			return 1
+		}
+		if len(cmdArgs) == 0 {
+			fmt.Fprintf(os.Stderr, "%s: missing command\n", progName())
+			return 1
+		}
```
omar-polo: is this redundant with the `if len(cmdArgs) == 0` just below?

technically a change with the current model, where `plakar` and `plakar at foo` gives two different outputs, but while we're here it's a nice simplification IMHO.

main.go
```diff
@@ -322,8 +327,20 @@ func entryPoint() int {
...
 	ctx.MaxConcurrency = opt_maxConcurrency
 
-	if flag.NArg() == 0 {
-		fmt.Fprintf(os.Stderr, "%s: a subcommand must be provided\n", filepath.Base(flag.CommandLine.Name()))
+	// An "at" with nothing behind it is a mistake of its own.
+	if hadAt {
+		if atRepository == "" {
+			fmt.Fprintf(os.Stderr, "%s: missing plakar repository\n", progName())
+			return 1
+		}
```
omar-polo, line 343: this is quite nice instead.  it's a change with the current behaviour, where `plakar at ""` will implicitly operate at `.`, but it's nice to catch this now!

omar-polo (approved):
nits, but overall very good, thanks ! =)

subcommands/cobra_test.go
poolpOrg: unsure what looks wrong to you ? the fact that you can --o ?

omar-polo: it's the fact that we're treating a simple option as a long one.  i fear this will bite us when we'll remove the compat.  just that =)

poolpOrg: By then, we will have redone the CLI options I believe, so let's not block on a conversion over future

omar-polo: okay!

cmd.go
poolpOrg: discussed with mathieu because he highlighted the same, will simplify later as we have more cleanup to do on the conversion once the mechanical changes in.

main.go
mathieu-plak: Fixed by yours truly.

subcommands/subcommands.go
poolpOrg: dropped comment

Issue #2303, opened 2026-08-06
[Bug]: backup to S3/B2 hangs indefinitely with zero I/O progress, never times out and leaves /tmp working dirs behind


pierrelcm:
### Describe the bug

Since 2026-07-30, every scheduled backup to a Backblaze B2 store (via the `s3` plugin) hangs indefinitely instead of completing. The process stays alive and keeps accumulating CPU time, but performs **no I/O at all** — `/proc/<pid>/io` counters are byte-identical when sampled 8 seconds apart. There is no error, no timeout, and nothing written to the log after the initial start line.

The same job had been running fine for months: the last successful snapshot completed in **1m28s** for a 4.8 GiB source. The hung run I caught live had been running for **9h48m**.

Two consequences make this worse than a simple stall:

1. **No timeout.** The run never gives up, so a daily timer produces a process that is still alive when the next one fires. On a 3.7 GiB machine it held ~1.2 GiB RSS and pushed the box into OOM-killing unrelated processes.
2. **`/tmp` working directories are never cleaned up.** Each hung run leaves a `/tmp/plakar-backup-<uuid>-<n>` directory behind. Five had accumulated over 8 days totalling 499 MB on a 1.9 GB tmpfs — which then broke unrelated software that needed `/tmp` space. This part looks related to #2254 and #1889, but here it is triggered by the hang rather than by a full cache.

Most importantly: **the backups silently stopped working.** Nothing failed loudly, the timer kept reporting activity, and the store simply stopped receiving snapshots for 8 days without any signal.

### Steps to reproduce

Not reproducible on demand — it started occurring consistently from a given day onward. Setup:

1. Store: Backblaze B2 via `s3@v1.1.2`
2. A systemd **user** service (`Type=oneshot`) triggered by a daily timer, running:
   ```
   plakar -quiet -concurrency 16 -keyfile <path> at @<store> backup -ignore-file <path> /home/<user>
   ```
3. Source is ~4.8 GiB on local ext4; the unit sets `MemoryHigh=1200M` and `Nice=10`.
4. Ran fine daily until 2026-07-29. From 2026-07-30 onward, **8 consecutive runs produced no snapshot at all** and never returned.

[...]

PR #2309, opened 2026-08-14
cached: unblock the state rebuild when the context is cancelled

ousamabenyounes:
`rebuildStateRequest` blocks in `dec.Decode()` waiting for cached's answer, and
nothing was watching the context there. A cancelled run stayed parked on that
read until the daemon replied, so the deferred cleanup never got to run.

A `context.AfterFunc` now closes the client connection on cancellation: the
pending `Decode` returns, `ctx.Err()` is reported to the caller, and the
deferred cleanup runs.

## Scope — issue #2303 is **not** solved by this

This was opened against #2303, and @mathieu-plak is right that it does not
resolve it: that issue wedges gRPC on the socket from both sides, which closing
the `cached` client cannot unstick. The closing link to that issue is removed
rather than left to imply something this PR does not deliver.

What remains is a standalone robustness fix in `cached`: a cancelled context
should not leave the caller blocked forever on a response that will never
arrive. Happy to close this if you would rather not carry an unrelated
hardening change — just say the word.

## Test verification (RED → GREEN)

RED — on unmodified `main`, with only the new test applied:

```
=== RUN   TestRebuildStateCancellationInterruptsResponseWait
    cached_test.go:380: state rebuild did not stop after context cancellation
--- FAIL: TestRebuildStateCancellationInterruptsResponseWait (1.00s)
FAIL	github.com/PlakarKorp/plakar/cached	1.010s
```

The fake server accepts the request and then never answers, so the RED run sits
on `Decode` until the test's own timeout fires.

GREEN — with the fix:

```
--- PASS: TestRebuildStateCancellationInterruptsResponseWait (0.00s)
ok  	github.com/PlakarKorp/plakar/cached	0.172s
```

## Full local suite

Replaying `.github/workflows/go.yml` and `go-windows.yml` locally:

```
go build -v ./...                                     ok
GOOS=windows GOARCH=amd64 go build -v .               ok
gofmt -l cached                                       clean
go vet ./...                                          clean
timeout -s QUIT 5m go test -v -p 4 -coverprofile=coverage.out \
  -covermode=atomic -timeout 2m ./...                 1115 passed, 0 failed
```

Coverage on the changed production lines of `cached/cached.go`: **100%**.

### DIFF
--- a/cached/cached.go
+++ b/cached/cached.go
@@ -1,6 +1,7 @@
 package cached
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -52,18 +53,24 @@ func rebuildStateRequest(ctx *appcontext.AppContext, req *RequestPkt) (int, erro
 	}
 	defer client.Close()
 
+	// Closing the connection is what makes the Decode below return: without
+	// it a cancelled context leaves us parked on a response that never comes.
+	// The returned func unregisters the hook once the request is done.
+	stopCloseOnCancel := context.AfterFunc(ctx, func() { _ = client.Close() })
+	defer stopCloseOnCancel()
+
 	if err := client.enc.Encode(req); err != nil {
 		return 1, err
 	}
 
 	response := &ResponsePkt{}
 	if err := client.dec.Decode(response); err != nil {
+		if err := ctx.Err(); err != nil {
+			return 1, err
+		}
 		// The server closed the connection without a response packet.
 		if err == io.EOF {
 			return 0, nil
-		}
-		if err := ctx.Err(); err != nil {
-			return 1, err
 		}
 		return 1, fmt.Errorf("failed to decode response: %w", err)
 	}
--- a/cached/cached_test.go
+++ b/cached/cached_test.go
@@ -67,6 +67,10 @@ type serverBehavior struct {
 
 	// onRequest, if set, receives every decoded request packet.
 	onRequest func(*RequestPkt)
+
+	// waitForClientClose leaves the request unanswered until the client closes
+	// the connection.
+	waitForClientClose bool
 }
 
 // startFakeServer listens on <cacheDir>/cached.sock and serves connections
@@ -151,6 +155,11 @@ func serveConn(conn net.Conn, b serverBehavior) {
 	}
 	if b.onRequest != nil {
 		b.onRequest(pkt)
+	}
+	if b.waitForClientClose {
+		var buf [1]byte
+		_, _ = conn.Read(buf[:])
+		return
 	}
 
 	if b.closeAfterRequest {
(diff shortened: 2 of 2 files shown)


cached/cached.go
```diff
@@ -52,19 +52,29 @@ func rebuildStateRequest(ctx *appcontext.AppContext, req *RequestPkt) (int, erro
...
 
+	stopWatching := make(chan struct{})
+	defer close(stopWatching)
+	go func() {
+		select {
+		case <-ctx.Done():
+			_ = client.Close()
+		case <-stopWatching:
+		}
+	}()
```
omar-polo: I don't understand this: you're creating a channel, close it when the function ends, and add it to an (empty) arm of a select.  what is this doing actually?

mathieu-plak:
Even if the fix made sense, this is not fixing the issue linked, that issue leads to GRPC being stuck on the socket on both sides so this would not save you.

ousamabenyounes:
Thanks both — two separate points, both taken.

**@omar-polo** — you were right to stop at that block: it was a hand-rolled `context.AfterFunc`, and reading it back your description of it ("an empty arm of a select") is exactly how it landed. The channel existed only to unregister the watcher when the function returned. That is what `context.AfterFunc` does in one call, so the goroutine, the channel and the select are gone:

```go
// Closing the connection is what makes the Decode below return: without
// it a cancelled context leaves us parked on a response that never comes.
// The returned func unregisters the hook once the request is done.
stopCloseOnCancel := context.AfterFunc(ctx, func() { _ = client.Close() })
defer stopCloseOnCancel()
```

**@mathieu-plak** — agreed, and thank you for catching it. Closing the `cached` client does nothing for a gRPC socket wedged from both ends, so `Fix #2303` was a claim this PR could not back. I have dropped the link from the description rather than leave it implying a fix it does not deliver.

That leaves a standalone robustness fix: `rebuildStateRequest` blocks in `dec.Decode()` with nothing watching the context, so a cancelled run stays parked there until the daemon answers and the deferred cleanup never runs. If you would rather not carry that as an unrelated hardening change, say so and I will close this — no argument from me.

Validation, in `b3f1a0f`: RED on unmodified `main` with only the new test applied — `TestRebuildStateCancellationInterruptsResponseWait` sits on `Decode` until its own timeout fires (`state rebuild did not stop after context cancellation`, 1.00s). GREEN with the fix (0.00s). Also rebased onto current `main`.

**One thing I cannot close out myself:** the Linux `build` job on `b3f1a0f` came back red, and I have not been able to identify what failed. The `Test` step pipes `go test -json` straight into `go-junit-report`, so the job log contains no test output at all — only `Process completed with exit code 1` — and the results only exist in the Codecov upload. Re-running the job needs write access I do not have on a fork PR.

What I can say is that it does not reproduce here. On `b3f1a0f`: the full suite three times (`1115 passed, 0 failed`), `cached` and `subcommands/pkg` five more times, and one replay of the exact CI pipeline (`-v -p 4 -covermode=atomic -timeout 2m -json | go tool go-junit-report -parser gojson -set-exit-code`) pinned to two cores to mimic the runner — `junit.xml` came back with zero `<failure>` elements. `go build -v ./...`, the windows cross-build, `gofmt -l` and `go vet ./...` are all clean, and the changed production lines are at 100% coverage.

I did hit `panic: pebble: closed` in `subcommands/pkg` (`TestPkgCreateExecuteFailsOnBadConnector`, inside kloset's `caching/pebble.(*cache).Scan`) on an unmodified `main` baseline run earlier today, and never again since — so a pre-existing flake is my best guess, but I would rather flag it than assert it. If someone with write access re-runs the job, that should settle it.

PR #2310, opened 2026-08-15
verify package signatures before installing

poolpOrg:
POC to be discussed.


signify/verifier.go
```diff
@@ -0,0 +1,130 @@
...
+	ErrNameMismatch   = errors.New("signature covers a different file")
+	ErrDigestMismatch = errors.New("package contents do not match the signed checksum")
+	ErrNoChecksum     = errors.New("signature carries no checksum")
+	ErrBadChecksum    = errors.New("unusable signed checksum")
+)
+
+// Re-exported so callers need not import the library directly.
+var ParseChecksums = gosignify.ParseChecksums
+
+type Verifier struct {
```
mathieu-plak, line 39: What's the purpose of this? It's always wrapped in an adapter. Drop the adapter and implement everything here no? Too much abstraction makes this hard to follow.

signify/adapter.go
```diff
@@ -0,0 +1,161 @@
...
+	if res != nil {
+		p.mu.Lock()
+		p.results[filename] = res
+		p.mu.Unlock()
+	}
+
+	return nil
+}
+
+func (p *PackageVerifier) Result(filename string) *Result {
```
mathieu-plak: I was curious as to how anyone could justify calling Verify concurrently with Result, well no one calls Result.

appcontext/appcontext.go
```diff
@@ -9,14 +9,16 @@ import (
...
 
 type AppContext struct {
 	*kcontext.KContext
 
-	cookies *cookies.Manager `msgpack:"-"`
-	pkgmgr  *pkg.Manager     `msgpack:"-"`
-	Config  *config.Config   `msgpack:"-"`
+	cookies     *cookies.Manager         `msgpack:"-"`
+	pkgmgr      *pkg.Manager             `msgpack:"-"`
+	pkgverifier *signify.PackageVerifier `msgpack:"-"`
```
mathieu-plak: I would have loved to avoid this. Maybe pkg manager should just hand over the verifier? Won't die on that hill, it's just food for thought.

signify/builtin.go
```diff
@@ -0,0 +1,44 @@
...
+		name: "plakar-20260815",
+		key: "untrusted comment: plakar signify public key\n" +
+			"RWT9B2Rj2tbKp89csIacV+QlvjDB4EHX9rAmS2pqgSXBbFovoLruzuSo\n",
+		scopes: []string{
+			// The parent of community/ and enterprise/, so one key
+			// covers the whole dist tree.
+			"https://plakar.io/dist/plugins/kloset/",
+
+			// A genuinely signed artifact is no less trustworthy
+			// for having been carried in by hand.
```
mathieu-plak, line 40: Not sure what that means sorry...

signify/builtin.go
```diff
@@ -0,0 +1,44 @@
...
+
+type builtinKey struct {
+	name   string
+	key    string // a signify public key file, verbatim
+	scopes []string
+}
+
+// A set rather than a single key so rotation works: a new key ships in one
+// release and signs from the next, while the outgoing one stays valid.
+// Selection is by key number, so it never surfaces to users. Names are dated
```
mathieu-plak: Not sure I understand the selection process. "by key number", so what current is always 0 and next release is 1?

signify/verifier.go
```diff
@@ -0,0 +1,130 @@
...
+var (
+	ErrUnsigned       = errors.New("package is not signed")
+	ErrNameMismatch   = errors.New("signature covers a different file")
+	ErrDigestMismatch = errors.New("package contents do not match the signed checksum")
+	ErrNoChecksum     = errors.New("signature carries no checksum")
+	ErrBadChecksum    = errors.New("unusable signed checksum")
+)
+
+// Re-exported so callers need not import the library directly.
+var ParseChecksums = gosignify.ParseChecksums
```
mathieu-plak: This is only called in this  file, and never exposed to any callers (I wouldn't expect it to be)

signify/verifier.go
```diff
@@ -0,0 +1,130 @@
...
+	// signature must be SHA256. Rejected here rather than left to the
+	// digest comparison, which would report a mismatch instead of the
+	// actual problem.
+	if !strings.EqualFold(want.Algorithm, "SHA256") {
+		return nil, fmt.Errorf("%w: %s is signed with %s, want SHA256",
+			ErrBadChecksum, filename, want.Algorithm)
+	}
+
+	// Hashed here rather than through want.Verify so the digest can be
+	// reported: a mismatch is far easier to diagnose with both values.
```
mathieu-plak: The actual library does in fact return an error with the right "mismatch" line.

pkg.go
```diff
@@ -29,13 +30,25 @@ func setupPkgManager(ctx *appcontext.AppContext, dataDir, cacheDir string) error
 		return fmt.Errorf("failed to init the package manager: %w", err)
 	}
 
+	// The trust store is built once per run from the keys compiled into
+	// this binary plus any the user added. A failure to read it is fatal
+	// rather than degraded: continuing with fewer keys than the user
+	// configured would silently install packages they never trusted.
```
mathieu-plak, line 36: This comment is misleading at best imho.

signify/scope.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, fmt.Errorf("%w: must not contain a query or fragment", ErrInvalidScope)
+	}
+
+	return &Scope{URL: normalizeURL(u)}, nil
+}
+
+// Resolves dot segments, which url.Parse does not: without it an origin like
+// /dist/plugins/kloset/../../../etc/ would satisfy a scope of
+// /dist/plugins/kloset/ while addressing /etc/. The trailing slash lets
+// matching compare whole segments.
```
mathieu-plak, line 77: Open question but wouldn't trusting just domain names be enough?

mathieu-plak (approved):

mathieu-plak (commented):
Failed to click.

signify/verifier.go
poolpOrg: you are right.

just for history, the initial plan was to sign any .sum file and GNU sha256sum doesn't output OpenBSD style tagged checksums, so I left the possibility to swap the checksums parsing function.

I then decided to stick to OpenBSD only format which GNU sha256sum can also emit, just not as default. I'll kill this indirection.

signify/builtin.go
poolpOrg: If I copy and offline package locally along with its sumsig, it's not less trustworthy that if I download it from dist directly (ie: in my dev env, doing a download of several ptar's along their sum.sign to a local directory and installing from there).

signify/scope.go
poolpOrg: fair question, but this completely kills the possibility to provide different signing keys for community vs enterprise and I'm unsure I want to prevent that in an irrevocable way at this point :-/

signify/verifier.go
poolpOrg: true, leftover from providing a custom parsing function for non-OpenBSD style sha256sum ... which I decided to go against ultimately.

pkg.go
poolpOrg: how so ?

poolpOrg: how so ?

signify/scope.go
poolpOrg: fair question, but doing so kills the possibility of using different keys for community and enterprise which I'm not sure I want to do at this point even if for the time being the same key is being used for both out of practicality.

signify/verifier.go
poolpOrg: removed

signify/builtin.go
poolpOrg: an artifact that has been genuinely signed using our key is as trustworthy if artifact + .sig are available in a local directory as if they had been downloaded from the dist url (ie: you doing a wget to a local directory in dev).

signify/scope.go
mathieu-plak: Good point on enterprise vs community.

mathieu-plak (approved):

poolpOrg:
@omar-polo ?

omar-polo (approved):

PR #2313, opened 2026-08-17
fix manual page installation

stspdotname:
The cmd/plakar directory doesn't exist anymore.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,8 +18,7 @@ install:
 	mkdir -p ${DESTDIR}${BINDIR}
 	mkdir -p ${DESTDIR}${MANDIR}/man1
 	${INSTALL_PROGRAM} plakar ${DESTDIR}${BINDIR}
-	find cmd/plakar -iname \*.1 -exec \
-		${INSTALL_MAN} {} ${DESTDIR}${MANDIR}/man1 \;
+	find . -name \*.1 -exec ${INSTALL_MAN} {} ${DESTDIR}${MANDIR}/man1 \;
 
 check: test
 test:
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,11 @@ plakar:
 
 install:
 	mkdir -p ${DESTDIR}${BINDIR}
-	mkdir -p ${DESTDIR}${MANDIR}/man1
+	mkdir -p ${DESTDIR}${MANDIR}/man{1,5,7}
 	${INSTALL_PROGRAM} plakar ${DESTDIR}${BINDIR}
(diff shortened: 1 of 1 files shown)


omar-polo (changes requested):
oops, good catch!

can you also fix the installation of `plakar-query.7` while at it?  thanks!

omar-polo (approved):
last nit: i'd probably spell the various mkdir for the section by themselves, I'm not sure any posix shell is required to implement the `{1,5,7}`-style expansion.

in any case, LGTM, thanks!

poolpOrg (approved):

QuentinVigand (approved):
Nice 👌 

Is there a "rule" as to when to add variables?

For example would you do that since it is repeated 4 times or not because it shouldn't be re-used anymore and is a bit less straightforward to read. 
```Makefile
MAN_INSTALL_PATH = ${DESTDIR}${MANDIR}
```

stspdotname:
> Nice 👌
> 
> Is there a "rule" as to when to add variables?
> 
> For example would you do that since it is repeated 4 times or not because it shouldn't be re-used anymore and is a bit less straightforward to read.
> 
> ```makefile
> MAN_INSTALL_PATH = ${DESTDIR}${MANDIR}
> ```

Can be done. I think it's just a matter of taste. I see no technical reasons for or against doing this.
In my experience, having $(DESTDIR) all over the place in Makefiles is quite common. I didn't even notice the repetition.

stspdotname:
> last nit: i'd probably spell the various mkdir for the section by themselves, I'm not sure any posix shell is required to implement the `{1,5,7}`-style expansion.
> 
> in any case, LGTM, thanks!

I must admit I don't know how common support for {1,5,7} is. Changing this seems trivial enough to just send the change instead of going through a PR?

omar-polo:
@QuentinVigand 

> Is there a "rule" as to when to add variables?
>
> For example would you do that since it is repeated 4 times or not because it shouldn't be re-used anymore and is a bit less straightforward to read.
>
> MAN_INSTALL_PATH = ${DESTDIR}${MANDIR}

I admit it's just the style I'm used; I've seen plenty of time it being written this way, from the *BSD ports tree to (some) third-party software.  Enough that I'm completely used to the names and their "joining".  But I won't oppose in any way some extra vars in case.

omar-polo:
@stspdotname fixed in https://github.com/PlakarKorp/plakar/pull/2321, which also includes the posix legalese quotes :D 

(i'm actually happy because i've learned something new today :D)

PR #2314, opened 2026-08-17
more api test cleanup

omar-polo:
drop some duplicate tests, then try to move them under a global top-level per-file test, so fixtures can be generated usually once.  This brings the regress from ~20s down to 3s without affecting coverage at all.  While here, also sneak in some nitpicks.

better read commit by commit with whitespaces disabled =)


poolpOrg (approved):

mathieu-plak (approved):
Nice! Thanks for doing that

PR #2316, opened 2026-08-17
action: force build of the tools

omar-polo:
this should at least make it more obvious if go fails at downloading the tools (or their deps) for the test step.


poolpOrg (approved):

mathieu-plak (approved):
Let's see..

PR #2317, opened 2026-08-17
main: generate the shell completion scripts

poolpOrg:
Builds on #2306, which turns the commands into a cobra tree; this makes
cobra generate the completion scripts from it.

    plakar completion bash    # also zsh, fish, powershell

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -1,0 +1,298 @@
+package main
+
+import (
+	"encoding/hex"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/PlakarKorp/kloset/caching"
+	"github.com/PlakarKorp/kloset/caching/pebble"
+	"github.com/PlakarKorp/kloset/connectors/storage"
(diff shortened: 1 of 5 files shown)


complete.go
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	if err := completionUnlock(ctx, serialized); err != nil {
+		closer()
+		return nil, nil, func() {}, err
+	}
+
+	// repository.New builds the state here rather than asking cached for
+	// it: spawning a daemon to answer a TAB is not worth it, and cached
+	// may not even be reachable.
```
mathieu-plak: Reword this (or just remove the comment entirely) it's very misleading for humans

omar-polo (approved):

mathieu-plak (approved):
LGTM. Thanks for dropping that unnecessary snapshot.Load. 

I'm curious to see how it behaves on remotes, 700ms is really not a lot (but a lot when you are hanging on a TAB-completion)

PR #2320, opened 2026-08-18
chore: .idea in .gitignore

QuentinVigand:
No IDE specific shared config, ignore Jetbrain's .idea directory.


mathieu-plak:
This repository does not use so called "semantic commit message tags". Otherwise OK

mathieu-plak (approved):

PR #2328, opened 2026-08-21
fix: [Bug]: mount has problems with windows drive letters

### LINKED ISSUE #2076: [Bug]: mount has problems with windows drive letters
### Describe the bug

Snapshots of paths containing Windows drive letters (C:) is not correctly rendered when mounted to a http endpoint.
If a specific snapshot is specified a different codepath is taken that skips the driveletter path and thus works as expected. If I do something similar to mount.go l:74-88 in http.go l:86-98 it works, with `subFS, err := fs.Sub(pvfs, path[1:])` being the difference maker.

Aligning the codepaths like above fixes functionality but really only works around a bug that may or may not be important to fix.

I have tried to find the root cause and I believe FileInfo.LName gets shortend to "/" somewhere at least close to vfs.go and entry.go in kloset, but I'm unable to find where. If I run in debug mode and set a break at vfs.go:174 and when entry.ParentPath is "/C:" I force entry.FileInfo.Lname to "C:" it all works, a "C:/" will be rendered, and clicking the link will allow me to drill down into the snapshot. 

Happy to submit a PR to align codepaths, and happy to get some hints on where to look for the root cause

### Steps to reproduce

On Windows run the following
1. plakar at repo create
2. plakar at repo backup 
3. plakar at repo backup 
4. plakar at repo ls
Two snapshots listed:
2026-04-09T06:05:28Z   4fd352a0   237 MiB        1s /C:/tmp
2026-04-09T06:05:22Z   a43049f5   182 MiB        1s /C:/tmp
5. plakar at repo mount -to http://localhost:5632
6. Open http://localhost:5632 in web browser. Shows the two snapshots as expected:

Image

7. Click on either of the snapshots and error occurs

Image
8. For reference plakar at repo mount -to http://localhost:5632 4fd352a0 works as expected

### Expected behavior

Driver letter rendered

Image

### Actual behavior

see above

### Plakar version

plakar/v1.1.0-beta.7

### OS, version and architecture

Windows 11

### Store location

local fs

### Backup source location

local fs

### Restore destination (if applicable)

_No response_

### Debug output

```shell

```

### Additional context

_No response_

Roshan931:
Fixes #2076

## Root cause

mount HTTP handler emits a scheme-ambiguous relative redirect for Windows drive-letter directories

## Changes

- NewDynamicSnapshotHandler now performs the missing trailing-slash redirect itself when the requested snapshot entry is a directory whose base name contains a colon, sending an absolute-path Location (/<snapshot>/C:/ plus any query) instead of letting net/http's localRedirect emit the ambiguous relative "C:/"; all other paths keep the stdlib behaviour. Regression tests cover the drive-letter redirect, query preservation, the unchanged colonless redirect, and serving the drive-letter subtree.

### DIFF
--- a/subcommands/mount/http/http.go
+++ b/subcommands/mount/http/http.go
@@ -148,7 +148,7 @@ func NewDynamicSnapshotHandler(list ListFn, open OpenFn) http.Handler {
 		}
 
 		p = strings.TrimPrefix(p, "/")
-		snap, _, _ := strings.Cut(p, "/")
+		snap, entrypath, _ := strings.Cut(p, "/")
 
 		snapFS, err := open(r.Context(), snap)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


dsobirk:
Thanks for trying to fix this. Unfortunately the problem seems to persist. As far as I can tell, the "offending" call to `absoluteDirRedirect` has an empty `entrypath` so will never make it to the `relativeRedirectParsesAsScheme` which I assume was the idea.

PR #2330, opened 2026-08-21
[OSS] Enable search and filtering on the integrations page

insanecoding:
Update the endpoint `/api/proxy/v1/integration` 
* add `search` capabilities (by name or display name)
* add filter by `installation_status: "installed" | "not-installed" | "upgradable"`
* add filter by `type: "storage" | "source" | "destination" | "provider"`

A cache from `go.omarpolo.com/ttlmap` was added to avoid re-fetching a remote catalog of 449 KB on every request

AI was used for this PR. It was guided to use the existing patterns and conventions, but it might still need a more careful review

### DIFF
--- a/api/api_proxy.go
+++ b/api/api_proxy.go
@@ -20,13 +20,7 @@ var SERVICES_ENDPOINT = "https://api.plakar.io"
 
 // integrationsCache memoizes pkg.Manager.Query() across requests so the
 // /integration list and detail endpoints don't fetch the ~449 KB remote
-// catalog from api.plakar.io on every hit. The handler runs all filters
-// (search, installation_status, type, tag) over the cached slice, so a
-// single entry under a constant key covers every call.
-//
-// TTL is short enough that catalog updates propagate without operator
-// action. install/uninstall explicitly invalidate the entry so
-// state-changing operations reflect immediately.
+// catalog from api.plakar.io on every hit.
 const integrationsCacheKey = "all"
 
 var integrationsCache = ttlmap.New[string, []*pkg.Integration](5 * time.Minute)
@@ -169,9 +163,6 @@ func (ui *uiserver) servicesSetAlertingServiceConfiguration(w http.ResponseWrite
 	return json.NewEncoder(w).Encode(alertConfig)
 }
 
-// integrationMatchesSearch reports whether the integration's Name or
-// DisplayName contains needle (case-insensitive substring). needle must
-// already be lowercased. An empty needle matches everything.
 func integrationMatchesSearch(it *pkg.Integration, needle string) bool {
 	if needle == "" {
 		return true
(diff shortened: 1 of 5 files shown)


brmzkw:
I'm unsure about the cache: what if an integration is added or removed outside of the API? (with the cmdline)

api/api_proxy.go
```diff
@@ -175,24 +186,34 @@ func (ui *uiserver) servicesGetIntegration(w http.ResponseWriter, r *http.Reques
...
-		if err != nil {
-			return err
+	if err != nil {
+		return err
+	}
+
+	var i int64
+	for _, integration := range integrations {
+		if !integrationMatchesSearch(integration, needle) {
+			continue
```
omar-polo, line 282: gut feeling: the query should be moved to pkg.git, it'd be nice to have in the cli as well.  not a blocker, and i'll add it on that side myself if you want.

api/api_proxy.go
```diff
@@ -188,6 +188,40 @@ func integrationMatchesInstallationStatus(it *pkg.Integration, filter string) bo
...
 }
 
+// integrationMatchesTypes reports whether the integration has at least
+// one of the requested Types bools set. filter values are the plakar
+// Types-struct field names in lowercase — "storage", "source",
+// "destination", "provider". Empty filter matches everything; unknown
+// values match nothing. Any-of semantics: an integration matches if it
+// satisfies ANY of the requested types (OR within the type filter; AND
+// with the other filters).
+func integrationMatchesTypes(it *pkg.Integration, filter []string) bool {
```
omar-polo, line 187: this also could be directly in pkg.git.

omar-polo (commented):
Hello!

I've missed this PR, sorry.  Quick review for now though.  Generally, looks good to me and if it works and the rest of the ui team is happy, it could even go in as-is for me.

There are some things that I think will need to be moved to pkg.git so that other consumers can benefit, but this doesn't need to be done immediately.

omar-polo:
ah, forgot to add: i sort of agree with @brmzkw, but i can also see a point for having a small cache there.  maybe we want to lower the TTL so that a user doesn' have to wait 5 minute to see reflected the status of `plakar pkg ...`?

edit: or even just start without and see how it's going?  is the UI going to call those endpoints more aggressively now?

api/api_proxy.go
```diff
@@ -7,14 +7,28 @@ import (
...
 )
 
 var SERVICES_ENDPOINT = "https://api.plakar.io"
 
+// integrationsCache memoizes pkg.Manager.Query() across requests so the
+// /integration list and detail endpoints don't fetch the ~449 KB remote
+// catalog from api.plakar.io on every hit.
+const integrationsCacheKey = "all"
+
+var integrationsCache = ttlmap.New[string, []*pkg.Integration](5 * time.Minute)
```
brmzkw: it's not a big deal, but I'd rather avoid having global variables and abusing init for this. I'll make a commit to move this to the uiserver struct

api/api_proxy.go
brmzkw: resolving the comment — @omar-polo told he'll create serpate issues to the package manager.

api/api_proxy.go
brmzkw: resolving this issue as you told me (on discord) you'll create issues on pkg.git

api/api_proxy.go
brmzkw: fixed

omar-polo (approved):

Issue #2338, opened 2026-08-25
[Bug]: backup: incremental cost scales with tree size, not churn


jlutran:
Follow-up to the discussion in #2303, where @mathieu-plak invited a separate
issue for the "growth" question, and possibly related to #2277 (state/cache
rebuild cost at scale).

## Environment

- plakar v1.1.4 (APT repo), s3 integration, fs importer v1.1.2, kloset v1.1.3
- Debian 13 (kernel 6.12.86), 8 threads Xeon D-1521, 32 GiB RAM, source on local ZFS, store on OVH Object Storage

## Observation

Nightly incremental of a 737 GiB / ~500k-file Nextcloud data tree with ~0.01% daily churn:

| run | wall | in: (read from store) | out: (written) |
|---|---|---|---|
| seed | 7h47m | 0 B | 599 GiB |
| night 1 | 3h02m | **799 MiB** | 87 MiB |
| night 2 | 4h03m | **797 MiB** | 86 MiB |
| night 3 | 4h13m | **793 MiB** | 429 MiB |

Note night 3 had ~5× the churn of the others with the same metadata read and wall time — the cost really is churn-independent.

The unchanged-file skip works (86 MiB uploaded for 737 GiB), but every run:

- reads back an essentially identical ~800 MiB from the store — the parent snapshot's VFS/btree, I believe — and this does not shrink between consecutive runs, so the local cache doesn't appear to absorb it;
- burns ~100 min of user CPU comparing the tree.

Smaller sources on the same kloset behave proportionally (a 5 GiB / ~9k-file tree reads ~9 MiB and takes ~1–4 min), i.e. **cost tracks tree size, not the amount of changed data** — the same shape @pierrelcm measured in #2303 (`GetSnapshot` once per existing snapshot at startup; `GetBlob(btree-node)` serialized at ~119 ms per call against the store).

## The local filesystem is not the bottleneck

Walking the same tree with `find -newermt <date>` (which lstats every entry,
like the importer does):

- cold caches: **8m20s** (real), 1m13s sys
- immediately re-run, warm: **19 s**

So even a fully cold local walk accounts for ~8 of the ~240 minutes. The
remainder is store round-trips + CPU inside plakar/kloset.

## Questions

- Is the per-run ~800 MiB parent-VFS read from the store expected, or should the local cache serve it? (Is there a cache size limit or setting I'm  hitting at this scale?)
[...]

PR #2348, opened 2026-08-26
backup: don't scan the source twice when nothing displays the summary

ousamabenyounes:
`backup` builds a second importer for the same source and walks the whole tree
through it just to produce the filesystem summary, on every run. Only the TUI
displays that summary and the JSON stream forwards it — the stdio renderer,
which `main` selects for `-quiet`, `-silent`, `-stdio`, `-trace` and every
non-interactive run, drops it. This records on the app context whether the
renderer consumes the summary and lets `backup` skip the pre-scan when it does
not; on a 30000-file source that halves the stat calls on the source, 60517 →
30310 `newfstatat`.

Refs #2338 but does not fix it: the store-side cost that issue reports (parent
VFS/btree blobs re-read per run) lives in `kloset`, not here.

### DIFF
--- a/appcontext/appcontext.go
+++ b/appcontext/appcontext.go
@@ -27,6 +27,10 @@ type AppContext struct {
 
 	Quiet  bool
 	Silent bool
+
+	// ProgressSummary is set when the selected renderer consumes fs.summary,
+	// which costs backup an extra scan of the source.
+	ProgressSummary bool
 }
 
 func NewAppContext() *AppContext {
--- a/main.go
+++ b/main.go
@@ -195,8 +195,11 @@ func entryPoint() int {
 	ctx := appcontext.NewAppContext()
 	defer ctx.Close()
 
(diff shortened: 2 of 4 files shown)


poolpOrg:
generally ok but can you please install the PlakarKorp/claude-standard plugin before doing submissions ? it will make claude adapt to our usual style as your PR has too many comments that comment over obvious things.

ousamabenyounes:
i'll add this configuration to my Agent workflow for Plakar! 
so happy to see this convention. 
It's the first time I found this on a repos (maybe I missed it on others) 
the AI agent convention is a real solution to a better collaboration.

hope you're doing well @poolpOrg  :) you and all plakar team

poolpOrg:
Thanks, hope you're doing well too.

Can you start a new session with the plugin and ask it to rework this PR using our standards ?

poolpOrg (approved):

mathieu-plak (approved):

omar-polo (approved):
nice, thank you!

PR #2360, opened 2026-09-01
all: Make usage of a plaintext repo explicit.

mathieu-plak:
* This adds an env var, required to exist to have any command work with a plaintext repo.

* This is a security strategy, because it makes it obvious if someone is attempting to override your repo with a plaintext config rather than silently pushing cleartext backups.

### DIFF
--- a/main.go
+++ b/main.go
@@ -441,6 +441,11 @@ func entryPoint() int {
 			return exitcodes.RepoIncompatible
 		}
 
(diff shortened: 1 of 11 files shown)


poolpOrg (approved):
small, I likey

omar-polo (approved):
```
mail -s 'RE plaintext' mathieu@plakar.io <<EOF
ok op@ but please run ./subcommands/sync-man.sh too before committing.

Thanks,
Omar Polo
EOF
```

PR #2365, opened 2026-09-04
Add web API for diag chunkmap

insanecoding:
Add a new `GET /api/snapshot/chunkmap` endpoint with `?path=` query params to mirror the CLI API `plakar at @mystore diag chunkmap snapId:path1 snapId2:path2 snapId3:path3`

This API will be used by a new OSS UI page displaying chunkmaps. Example usage:
Screenshot 2026-09-07 at 10 57 55

Extra careful consideration might be needed regarding memory usage on heavy files. (But the PR reuses the CLI implementation for that, so nothing new is introduced)

(Created with AI, using PlakarKorp Standards)


poolpOrg (approved):
OK

PR #2366, opened 2026-09-05
config: support repository passphrase env

### LINKED ISSUE #1263: Environnment variable for destination repository
Hello

I can't use a plakar sync to sftp in a script, because it's waiting for the destination repository passphrase. I talk about the repository passord, not the ssh key passphrase. I can use the var PLAKAR_PASSPHRASE for the local repo, but I cannot see in the documentation any variable for the distant repository ?

ousamabenyounes:
Fix #1263.

Adds PLAKAR_REPOSITORIES_<NAME>_PASSPHRASE and _PASSPHRASE_CMD for named repository configs, with lookup kept separate from storage backend configuration so secrets are not persisted or passed through connector maps. This covers the main, completion, sync, and ptar repository-open paths, including ambiguous-name rejection when an exact env override is present.

## Test verification (RED to GREEN)
RED: with the production patch reverted, focused config/main tests failed to compile because the new repository passphrase APIs/signature were absent; the ptar regression entered the old noninteractive prompt loop and was interrupted.
GREEN: focused config/main/ptar regression tests passed, and the full local suite passed via ./run-ci.sh.

### DIFF
--- a/complete.go
+++ b/complete.go
@@ -191,9 +191,17 @@ func completionRepository() (*appcontext.AppContext, *repository.Repository, fun
 	ctx.SetCache(caching.NewManager(pebble.Constructor(cacheDir)))
 	nop = func() { ctx.GetCache().Close(); ctx.Close() }
 
-	storeConfig, err := ctx.Config.GetRepository(completionRepositoryPath(ctx))
+	repositoryPath := completionRepositoryPath(ctx)
+	storeConfig, err := ctx.Config.GetRepository(repositoryPath)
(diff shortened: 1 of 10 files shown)


omar-polo:
Hello,

Sorry for the long delay.  I don't think though that we want this in.  Nowadays we have `-o` options that can be passed for backup and restore, and we can have something similar for the store as well.  in addition, i'm not sure on having the _CMD variant that executes code out of an env variable...

PR #2369, opened 2026-09-09
Throttling parameters

mathieu-plak:
This enable usage of read_rate and write_rate configuration keys to limit bandwidth used by i/o.

### DIFF
--- a/main.go
+++ b/main.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"maps"
 	"os"
 	"os/signal"
 	"os/user"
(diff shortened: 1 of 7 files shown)


omar-polo (commented):
this is a bit more ewww IMHO.  i see two issues:

* as long as only `passphrase` is "special" it's fine to just have it with the other store options, but now we're overloading the store configuration with even repository-level stuff.

i didn't like it back then, but maybe we should follow @ericfaurot guts and split the repository settings (passphrase, now throttling) with the store config (the various params)

* if you really want for now to keep these in the configuration (which would be fair, since it's outside of this scope to introduce a different configuration section), you should plug ptar as well, otherwise in some cases those parameters are passed thru, and in others don't.

PR #2371, opened 2026-09-10
login: harden sign-ins with a completion code

poolpOrg:
* Makes phishing a login link much harder: the api (>= v1.0.17) shows a short code on the post-sign-in success page and only releases the token to the client that presents it. Whoever completes the sign-in hands the code to the device that started it, confirming both ends belong to the same person.

* plakar login now requests the code (github and email alike): when the api echoes the opt-in, the poll prompts for the code shown in the browser and retries with the X-Completion-Code header. A mistyped code simply re-prompts.

* The poll answers 401 for two reasons — a missing completion code (recoverable: prompt and retry) and a poll-secret mismatch (fatal). It tells them apart on the api's stable "code" field (completion_code_required / poll_secret_required, api >= v1.0.19), falling back to the error prose for older apis.

* An older api ignores the field and never echoes it, so the flow behaves as before against it. RunUI (the local ui's flow) is untouched.

### DIFF
--- a/login/login.go
+++ b/login/login.go
@@ -127,8 +127,12 @@ func (flow *loginFlow) poll(pollID string, iterations int, delay time.Duration, 
 	return "", fmt.Errorf("could not obtain token after %d iterations", iterations)
 }
 
+// maxErrorBodySize bounds how much of an error response body is read;
+// the expected content is a small JSON error envelope.
+const maxErrorBodySize = 4096
+
 func isCompletionCodeRequired(body io.Reader) bool {
-	data, _ := io.ReadAll(io.LimitReader(body, 4096))
+	data, _ := io.ReadAll(io.LimitReader(body, maxErrorBodySize))
 	var env struct{ Code string }
 	if json.Unmarshal(data, &env) == nil && env.Code != "" {
 		return env.Code == "completion_code_required"
--- a/login/login.go
+++ b/login/login.go
@@ -17,6 +17,7 @@
 package login
 
 import (
+	"bufio"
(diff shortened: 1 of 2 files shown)


login/login.go
```diff
@@ -104,6 +127,15 @@ func (flow *loginFlow) Poll(pollID string, iterations int, delay time.Duration,
 	return "", fmt.Errorf("could not obtain token after %d iterations", iterations)
 }
 
+func isCompletionCodeRequired(body io.Reader) bool {
+	data, _ := io.ReadAll(body)
```
omar-polo: a bit stupid, but to avoid being flagged by bots again, can you wrap this in a LimitedReader?

login/login.go
```diff
@@ -117,7 +149,13 @@ func (flow *loginFlow) Run(provider string, parameters map[string]string) (strin
 		return "", fmt.Errorf("unsupported provider: %s", provider)
 	}
 
-	if bodyBytes, err := json.Marshal(parameters); err != nil {
+	// parameters is map[string]string; completion_code is a JSON bool, so the
+	// request body is an any-valued copy with the flag added.
+	payload := map[string]any{"completion_code": true}
+	for k, v := range parameters {
+		payload[k] = v
+	}
```
omar-polo, line 161: maps.Copy?

omar-polo (approved):

login/login.go
poolpOrg: no they dont have the same key/value types

mathieu-plak (approved):

PR #2372, opened 2026-09-13
subcommands/config: drop http import feature

omar-polo:
was never documented anywhere and it's quite risky to blindly read a configuration from the network.


omar-polo:
maybe even want to add a flag to allow importing configuration that can execute commands?

poolpOrg (approved):

mathieu-plak (approved):

